PR #16, opened 2021-12-06
global: upgrade to v7.0.0dev5

ppanero:
Small summary, we should ticketize issues once we merge.

**Frontpage works properly**

Screenshot 2021-12-06 at 14 03 25

**Login/Singup** working well except the banner not being removed when clicked on the `x`
Screenshot 2021-12-06 at 14 03 39

**Deposit form** and other places e.g. communities or profile have issues with header heigth. Banners display with space in between and also the logo is bigger than it should

Screenshot 2021-12-06 at 14 05 06
Screenshot 2021-12-06 at 14 05 21
Screenshot 2021-12-06 at 14 04 31
Screenshot 2021-12-06 at 14 04 19


.invenio
```diff
@@ -5,13 +5,13 @@ logfile = /logs/invenio-cli.log
...
 project_shortname = zenodo
-project_site = zenodo.org
+project_site = zenodo-rdm.web.cern.ch
 github_repo = zenodo/zenodo-rdm
 description = Invenio RDM Zenodo Instance
 author_name = CERN
 author_email = info@zenodo.org
 year = 2020
-python_version = 3.6
+python_version = 3.8
```
slint, line 14: I'm getting some PTSD here, but would it be ok to use 3.9? I can imagine by the time we migrate, it would be best to be on the latest supported Python version

README.md
```diff
@@ -6,3 +6,35 @@
...
+# Invenio-OAuthclient
+# secrets will be injected on deployment
+CERN_APP_CREDENTIALS = {
+    "consumer_key": "CHANGE ME",
+    "consumer_secret": "CHANGE ME",
+}
+ORCID_APP_CREDENTIALS = {
+    "consumer_key": "CHANGE ME",
+    "consumer_secret": "CHANGE ME",
+}
```
slint, line 39: In general, for Zenodo we'll have the following OAuth methods:

- ORCiD
- GitHub
- OpenAIRE AAI (depends on https://github.com/inveniosoftware/invenio-oauthclient/pull/266)

slint (commented):
First quick look, some minor things

.invenio
ppanero: I agree, but atm not many people have tested/used it with 3.9 and there have been reports in discord about issues... Im just being safe for the two weeks we got 👼 , thats why I bumped only to 3.8

.invenio
```diff
@@ -3,15 +3,15 @@ flavour = RDM
 logfile = /logs/invenio-cli.log
 
 [cookiecutter]
-project_name = Zenodo
-project_shortname = zenodo
-project_site = zenodo.org
+project_name = Zenodo RDM
+project_shortname = zenodo-rdm
```
ppanero, line 7: This is required because othewise we are forced to have the project in a `zenodo` folder, which might collide with the current zenodo.

A more future proof option would be to "hardcode" services' names with the `container_name` attribute. This can be calculated based on the `project_shortname` at cookiecutting time. Currently, we are doing maths with compose version + hardcoded string (i.e. `<prefix_based_on_version>+_web_ui_1`). If the idea is approved I can work on it (cc @slint).

Issue #56, opened 2022-10-13
Create a Zenodo profile for describo


alejandromumo:
The goal of this issue is to create a describo profile to describe a deposit in InvenioRDM. 

The agreed fields for this iteration are the following:

- title
- author
- license
- keywords (subjects)
- datePublished

--- 

Tasks: 

- [x] create a profile compliant with Describo rules.

- [x] serve the file as a static file in `zenodo-rdm`.

--- 

Useful links:

- https://github.com/Arkisto-Platform/describo-online/wiki/Writing-a-domain-specific-profile
- https://github.com/UTS-eResearch/describo/blob/master/src/components/profiles/paradisec.describo.profile.json
- https://schema.org/ (property definitions)

PR #57, opened 2022-10-13
profile: added describo profile.

alejandromumo:
closes https://github.com/zenodo/zenodo-rdm/issues/56

### DIFF
--- a/static/describo/zenodo-describo-profile.json
+++ b/static/describo/zenodo-describo-profile.json
@@ -1,0 +1,133 @@
+{
+    "metadata": {
+        "name": "Zenodo Describo Profile",
+        "version": 0.1,
+        "description": "A profile for InvenioRDM reposiory",
+        "warnMissingProperty": true
+    },
+    "layouts": {
+        "Dataset": [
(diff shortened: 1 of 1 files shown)


static/describo/zenodo-describo-profile.json
```diff
@@ -0,0 +1,133 @@
...
+                },
+                {
+                    "id": "https://schema.org/Date",
+                    "definition": "inherit",
+                    "name": "datePublished",
+                    "label": "Publication Date",
+                    "type": [
+                        "Date"
+                    ],
+                    "required": false,
```
ppanero, line 51: This is required in RDM, just that the deposit form has a "right now" as default value. But there is no default at Marshmallow level, see [here](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/schemas/metadata.py#L343). So it will fail if not sent. Can Describo provide a "default" as we do in the deposit form?

alejandromumo: Defaults are not available for describo profiles. Perhaps`required` can be used here

ppanero (approved):

zzacharo (approved):

PR #61, opened 2022-10-19
Add custom code for support

zzacharo:
* address https://github.com/zenodo/zenodo-rdm/pull/71


site/setup.cfg
```diff
@@ -22,6 +22,10 @@ zip_safe = False
 [options.extras_require]
 tests =
     pytest-invenio>=2.1.0,<3.0.0
+    pytest-black>=0.3.0
+    sphinx>=4.5
+    invenio-access>=1.0.0,<2.0.0
```
alejandromumo, line 27: Not used right now, but it will be needed when testing endpoints.

site/zenodo_rdm/__init__.py
```diff
@@ -0,0 +1 @@
+__version__ = '0.0.0'
```
alejandromumo, line 1: Added, otherwise test for version was failing

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/CategoryDropdown.js
```diff
@@ -0,0 +1,69 @@
...
+            <>
+                <SelectField
+                    options={serializedCategories}
+                    fieldPath='category'
+                    label='Category'
+                    required
+                    width={4}
+                    onChange={this.onChange}
+                />
+                {this.state.activeCategory && <div dangerouslySetInnerHTML={{ __html: this.state.activeCategory.description }}/>}
```
alejandromumo, line 53: `category.description` comes from the backend (`config.py`) and it has some `html` embedded. 

Do we have alternate ways of doing this?

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
```diff
@@ -0,0 +1,290 @@
...
+import { TextField, TextAreaField, ToggleField } from "react-invenio-forms";
+import CategoryDropdown from "./CategoryDropdown";
+import * as Yup from "yup";
+import axios from "axios";
+import Dropzone from "react-dropzone";
+import FileTable from "./FileTable";
+import _remove from "lodash/remove";
+import { humanReadableBytes } from "react-invenio-deposit";
+
+// TODO translations
```
alejandromumo, line 14: TODOS are to be tackled in separate tasks. This PR is already too long

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
```diff
@@ -0,0 +1,290 @@
...
+    xsrfHeaderName: "X-CSRFToken",
+    headers: {
+        "Accept": "application/json",
+        'Content-Type': 'multipart/form-data',
+    },
+};
+
+const http = axios.create(apiConfig);
+
+function formikToFormData(values) {
```
alejandromumo, line 31: Even though `Formik` stores files, they were not being sent properly to the backend. Thus, when posting the form data I had to use `FormData`. 

If there is a cleaner / more correct way to do it, please let me know.

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
```diff
@@ -0,0 +1,290 @@
...
+
+const FileUploader = ({ dropzoneParams, maxFileSize, name, currentFiles, handleDelete }) => {
+    return (
+        <Dropzone {...dropzoneParams}>
+            {({ getRootProps, getInputProps }) => {
+                const inpProps = {
+                    ...getInputProps(),
+                    // Display the dropzone input, otherwise it's hidden by default.
+                    style: {
+                        'display': 'block'
```
alejandromumo, line 57: `react-dropzone` has an `input` element that is hidden. 

In other parts of invenio we keep it that way and re-implement it with either a button or a box for drag and drop. Here I went for a simpler solution, might not be the best though.

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
```diff
@@ -0,0 +1,290 @@
...
+                `Field must be at least ${this.props.descriptionMinLength} characters long.`
+            ).max(
+                this.props.descriptionMaxLength,
+                `Field must be at most ${this.props.descriptionMaxLength} characters long.`)
+                .required(),
+            sysInfo: Yup.boolean()
+        });
+
+        this.state = {
+            files: []
```
alejandromumo, line 108: Files are only validated by `react-dropzone` and no error/validation message is shown. Needs to be implemented.

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
```diff
@@ -0,0 +1,290 @@
...
+        const formData = formikToFormData(values);
+        try {
+            response = await http.post(apiEndpoint, formData);
+        } catch (error) {
+            let errMessage = error.message;
+            if (response) {
+                errMessage = error.response.data;
+            }
+            
+            // TODO show error notification
```
alejandromumo, line 124: Currently the form does not provide a lot of feedback to the user, both in error and success scenarios.

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
```diff
@@ -0,0 +1,290 @@
...
+                validateOnBlur={false}
+            >
+                {({ handleSubmit, setFieldValue }) => {
+                    let dropzoneParams = {
+                        onDropAccepted: (acceptedFiles) => {
+                            this.updateFiles(acceptedFiles, setFieldValue)
+                        },
+                        onDropRejected: (rejectedFiles) => {
+                            // TODO: show error message when files are rejected e.g size limit
+                            console.error(rejectedFiles[0].errors);
```
alejandromumo, line 177: Needs to be implemented: when the file is rejected by `react-dropzone`, we should somehow notify the user.

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
```diff
@@ -0,0 +1,290 @@
...
+        } else {
+            formData.append(key, currentValue);
+        }
+    })
+
+    return formData;
+}
+
+
+const FileUploader = ({ dropzoneParams, maxFileSize, name, currentFiles, handleDelete }) => {
```
alejandromumo, line 49: Perhaps could be moved to another file

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
```diff
@@ -0,0 +1,290 @@
...
+                                </label>
+                                <FileUploader
+                                    dropzoneParams={dropzoneParams}
+                                    maxFileSize={maxFileSize}
+                                    name='files'
+                                    currentFiles={this.state.files}
+                                    handleDelete={(file) => {
+                                        const newFiles = this.state.files;
+                                        _remove(newFiles, file);
+                                        this.updateFiles(newFiles, setFieldValue)
```
alejandromumo, line 248: File is removed from the component state and Formik. The component state is passed to the table for files listing. Formik is used to create the request object to post the form.

Note that `react-dropzone` is not being updated, I did not find a way of doing it.

site/zenodo_rdm/support/email.py
```diff
@@ -0,0 +1,100 @@
...
+
+class SupportEmailService(EmailService):
+    """_summary_
+    Provides an interface to send support related emails.
+    """
+    def __init__(self):
+        """Constructor."""
+        super().__init__()
+
+    def send_support_email(
```
alejandromumo, line 42: Perhaps this can be simplified into a method that works solely with contexts, as suggested by @slint personally.

site/zenodo_rdm/support/errors.py
```diff
@@ -0,0 +1,20 @@
...
+
+class SupportEmailNotSent(ZenodoRDMError):
+    def __init__(self, message):
+        """Initialise error."""
+        super().__init__(
+                "Support email failed to send. Error: {}".format(message)
+        )
+
+
+class FormValidationError(ZenodoRDMError):
```
alejandromumo, line 19: Currently not in use. Can be used when Marshmallow fails to validate the form data  or simply removed (marshmallow already throws a `ValidationError`) .

site/zenodo_rdm/support/support.py
```diff
@@ -0,0 +1,120 @@
...
+
+        try:
+            self.handle_form(data)
+        except (Exception) as e:
+            return str(e), 500
+
+        return redirect(url_for("invenio_app_rdm.index"))
+
+    def handle_form(self, form_data):
+        """Form controller."""
```
alejandromumo, line 56: Currently the only handler is to send two emails. However it can grow into different behaviors (e.g. based on a category do something else)

site/zenodo_rdm/support/support.py
```diff
@@ -0,0 +1,120 @@
...
+        )
+
+    def post(self):
+        """Receives a form, validates its data and handles it."""
+        data = {**request.form.to_dict(), "files": request.files.getlist("files")}
+
+        try:
+            data = self.validate_form(data)
+        except (ValidationError) as e:
+            return str(e), 400
```
alejandromumo, line 46: I am unsure whether this error handling is ideal.

Please let me know if it can be improved.

site/zenodo_rdm/templates/semantic-ui/zenodo_rdm/support.html
```diff
@@ -0,0 +1,52 @@
...
+{%- set user = current_user.is_authenticated and current_user %}
+{%- set userData = user and { 'email': user.email, 'full_name': user.user_profile.full_name }%}
+
+{% block first_section %}
+    <div id="contact-page" class="ui one column grid container">
+        <div class="row">
+            <div class="column">
+                <div class="panel-body mt-5">
+                    <h2 class="ui header mb-10">Contact us</h2>
+                    <p>Before making a request, you can have a look at our <a href="http://help.zenodo.org">FAQ</a> and <a href="http://about.zenodo.org">other resources</a> for more detailed information about the operation, features and goals of Zenodo.</p>
```
alejandromumo, line 16: links are still pointing to `zenodo.org`. Could be passed by config.

However many links in `zenodo-rdm` are currently pointing to `zenodo.org`

Dockerfile
```diff
@@ -10,6 +10,7 @@
 
 FROM inveniosoftware/centos8-python:3.8
 
+COPY site ./site
```
kpsherva, line 13: 👍

assets/less/zenodo-rdm/globals/site.variables
```diff
@@ -26,3 +26,5 @@
 @bold: 700;
 
 @headerFontWeight: @semibold;
+
+@cover-panel-padding: 40px;
```
kpsherva, line 30:
```suggestion
@coverPanelPadding: 2em;
```
or use 2*@defaultPadding

assets/less/zenodo-rdm/globals/site.overrides
```diff
@@ -38,4 +38,22 @@ body {
...
     margin-left: 0.3em;
   }
-}
\ No newline at end of file
+}
+
+#contact-page {   
+
+  .ui.container {
+    padding: 15px
```
kpsherva, line 46: there is builtin class in the theme
`p-15` no need to define a new one

assets/less/zenodo-rdm/globals/site.overrides
```diff
@@ -38,4 +38,22 @@ body {
...
+  }
+
+  .content-tab {
+    margin-bottom: 10px;
```
kpsherva, line 50: you could use `mb-10` instead

site/zenodo_rdm/__init__.py
kpsherva, line 1:
```suggestion
__version__ = '0.1.0'
```

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/CategoryDropdown.js
```diff
@@ -0,0 +1,69 @@
...
+import _find from 'lodash';
+
+class CategoryDropdown extends Component {
+
+    constructor(props) {
+        super(props);
+        let activeCategory = null;
+        if (Array.isArray(props.categories)) {
+            activeCategory = this.props.categories.find((el) => (el.key === props.defaultCategory))
+        }
```
kpsherva, line 13:
```suggestion
        super(props);
        let activeCategory;
        const { categories, defaultCategory } = props;
        if (Array.isArray(categories)) {
            activeCategory = categories.find((el) => (el.key === defaultCategory))
        }
```

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/CategoryDropdown.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    onChange = ({ data, formikProps }) => {
+        const activeCategory = this.props.categories.find((el) => (el.key === data.value))
+        this.setState({
+            activeCategory: activeCategory
+        });
+        formikProps.form.setFieldValue('category', data.value);
+    }
```
kpsherva, line 36:
```suggestion
        const { categories } = this.props;
        const activeCategory = categories.find((el) => (el.key === data.value))
        
        this.setState({
            activeCategory: activeCategory
        });
        formikProps.form.setFieldValue('category', data.value);
    }
```

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/CategoryDropdown.js
```diff
@@ -0,0 +1,69 @@
...
+
+    render() {
+        const { categories } = this.props;
+
+        const serializedCategories = Array.isArray(categories) ? categories.map((cat) => (this.serializeCategory(cat))) : []
```
kpsherva, line 41:
```suggestion
        const serializedCategories = categories.map((cat) => (this.serializeCategory(cat))) 
```
no need to check if it is array, the prop types are supposed to type check

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/CategoryDropdown.js
```diff
@@ -0,0 +1,69 @@
...
+    onChange = ({ data, formikProps }) => {
+        const activeCategory = this.props.categories.find((el) => (el.key === data.value))
+        this.setState({
+            activeCategory: activeCategory
+        });
+        formikProps.form.setFieldValue('category', data.value);
+    }
+
+    render() {
+        const { categories } = this.props;
```
kpsherva, line 39:
```suggestion
        const { categories } = this.props;
        const { activeCategory} = this.state;
```

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/CategoryDropdown.js
kpsherva, line 53:
```suggestion
                {activeCategory && <div dangerouslySetInnerHTML={{ __html: activeCategory.description }}/>}
```

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/FileTable.js
```diff
@@ -0,0 +1,95 @@
...
+    Table,
+} from "semantic-ui-react";
+import { humanReadableBytes } from "react-invenio-deposit";
+
+
+const FileTableRow = ({ handleDelete, file, decimalSizeDisplay }) => {
+    return (
+        <Table.Row key={file.name}>
+            <Table.Cell data-label={"Filename"} width={10}>
+                <div>
```
kpsherva, line 16:
```suggestion
                <>
```
can be replaced by react fragment, no div needed

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/FileTable.js
```diff
@@ -0,0 +1,95 @@
...
+    Popup,
+    Table,
+} from "semantic-ui-react";
+import { humanReadableBytes } from "react-invenio-deposit";
+
+
+const FileTableRow = ({ handleDelete, file, decimalSizeDisplay }) => {
+    return (
+        <Table.Row key={file.name}>
+            <Table.Cell data-label={"Filename"} width={10}>
```
kpsherva, line 15:
```suggestion
            <Table.Cell data-label="Filename" width={10}>
```

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/FileTable.js
```diff
@@ -0,0 +1,95 @@
...
+                <div>
+                    <label>{file.name}</label>
+                    <br />
+                    {file.checksum && (
+                        <div className="ui text-muted">
+                            <span style={{ fontSize: "10px" }}>{file.checksum}</span>{" "}
```
kpsherva, line 21: inline styling is violating CSP, could you move it to CSS intead?

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
```diff
@@ -0,0 +1,290 @@
...
+
+
+const FileUploader = ({ dropzoneParams, maxFileSize, name, currentFiles, handleDelete }) => {
+    return (
+        <Dropzone {...dropzoneParams}>
+            {({ getRootProps, getInputProps }) => {
+                const inpProps = {
+                    ...getInputProps(),
+                    // Display the dropzone input, otherwise it's hidden by default.
+                    style: {
```
kpsherva, line 56: can this styling be moved to theme files?

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
```diff
@@ -0,0 +1,290 @@
...
+                        'display': 'block'
+                    },
+                    name: name
+                }
+                return (
+                    <div {...getRootProps()}>
+                        <input {...inpProps} />
+                        <label className="helptext mt-0">
+                            {`Optional. Max attachments size:`} {humanReadableBytes(maxFileSize, false)}
```
kpsherva, line 65:
```suggestion
                            {`Optional. Max attachments size: ${humanReadableBytes(maxFileSize, false)`}
```

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
```diff
@@ -0,0 +1,290 @@
...
+    maxFileSize: PropTypes.number.isRequired,
+    name: PropTypes.string.isRequired,
+    currentFiles: PropTypes.array.isRequired
+};
+
+class SupportForm extends Component {
+
+    constructor(props) {
+        super(props);
+        this.FormSchema = Yup.object({
```
kpsherva, line 92: would it make sense to use backend validation like in the deposit form? I find yup syntax extremely unreadable

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
```diff
@@ -0,0 +1,290 @@
...
+        const { apiEndpoint } = this.props;
+        let response;
+        const formData = formikToFormData(values);
+        try {
+            response = await http.post(apiEndpoint, formData);
+        } catch (error) {
+            let errMessage = error.message;
+            if (response) {
+                errMessage = error.response.data;
+            }
```
kpsherva, line 122:
```suggestion
    onSubmit = async (values, formikBag) => {
        const { apiEndpoint } = this.props;

        const formData = formikToFormData(values);
        try {
            const response = await http.post(apiEndpoint, formData);
            window.location = response.request.responseURL;
        } catch (error) {
            let errMessage = error.message;
            if (response) {
                errMessage = error.response.data;
            }
```

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
```diff
@@ -0,0 +1,290 @@
...
+            
+            // TODO show error notification
+            console.error(errMessage); 
+        }
+
+        // TODO show a success notification
+        window.location = response.request.responseURL;
```
kpsherva, line 129:
```suggestion

```
moved above inside catch block

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
```diff
@@ -0,0 +1,290 @@
...
+
+        const defaultCategory = categories.length > 0 ? categories[0].key : '';
+
+        const initialValues = {
+            email: userMail,
+            name: name,
+            category: defaultCategory,
+            sysInfo: false,
+            files: null
+        }
```
kpsherva, line 158: does it make sense to be moved to the constructor instead of redeclaring it on each render?

site/zenodo_rdm/support/support.py
kpsherva: should raise a rest error, there are defined already in flask resources

site/zenodo_rdm/support/support.py
```diff
@@ -0,0 +1,120 @@
...
+
+        try:
+            self.handle_form(data)
+        except (Exception) as e:
+            return str(e), 500
```
kpsherva, line 51: same as above

site/zenodo_rdm/support/support.py
```diff
@@ -0,0 +1,120 @@
...
+                title="[{}]: {}".format(data.get('category'), data.get('subject'))
+            )
+        except (smtplib.SMTPSenderRefused) as e:
+            raise SupportEmailNotSent(
+                "There was an issue sending an email to the provided "
+                "address (ours), please make sure it is correct. "
+                "If this issue persists you can send "
+                "us an email directly to {}".format(self.email_service.support_emails)
+            )
+        except (Exception) as e:
```
kpsherva, line 84: this will catch any exception and will make debugging harder, is it on purpose ?

site/zenodo_rdm/support/support.py
```diff
@@ -0,0 +1,120 @@
...
+        try:
+            self.email_service.send_confirmation_email(recipients=data.get("email"))
+        except (smtplib.SMTPSenderRefused) as e:
+            raise ConfirmationEmailNotSent(
+                "There was an issue sending a confirmation email to the provided "
+                "address (yours), please make sure it is correct. "
+                "If this issue persists you can send "
+                "us an email directly to {}".format(self.email_service.support_emails)
+            )
+        except (Exception) as e:
```
kpsherva, line 103: same as above

site/zenodo_rdm/support/utils.py
```diff
@@ -0,0 +1,15 @@
...
+    template = current_app.jinja_env.get_or_select_template(template)
+    return template.render(context)
+
+
+def format_user_email(email, name):
+    """Format the user's email as 'Full Name <email>' or 'email'."""
+    if name:
+        email = "{name} <{email}>".format(name=name, email=email)
+    # Remove commas (',') since they mess with the "TO" email field
+    return email.replace(",", "")
```
kpsherva, line 15: what if there are several email addresses passed in the email field? can this happen?

site/zenodo_rdm/templates/semantic-ui/zenodo_rdm/support.html
kpsherva: we agreed to keep it for now

kpsherva (commented):
all the comments can be addressed in the second iteration if that is easier to manage

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/CategoryDropdown.js
zzacharo: minor: this can be simplified to 
```suggestion
        ...
        this.setState({activeCategory});
        ...    
```

assets/less/zenodo-rdm/globals/site.overrides
alejandromumo: This class is not used anymore, it was a leftover from the migration.

site/tests/support/conftest.py
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2022 CERN.
+#
+# invenio-administration is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
```
ppanero, line 7: Is this file needed? If no fixtures are setup, the conftest is not needed

site/tests/support/test_contact_form.py
```diff
@@ -0,0 +1,91 @@
...
+        "email": "john_doe@invenio.org",
+        "category": "file-modification",
+        "description": "Hi! I need some help with my record ...",
+        "subject": "Record file is wrong",
+        "sysInfo": False,
+        "files": [],
+    }
+
+
+def test_endpoints():
```
ppanero, line 39: Should we clean unimplemented tests and make them an issue? I fear otherwise it will just be leftover code and we will forget about it

site/zenodo_rdm/__init__.py
ppanero: Maybe this is a question for @zzacharo, if the package is only accessible inside the instance (no pypi) and is installed with editable. what is the idea behind having a version string? i.e. the version string will never be used for installing since we wont be able to install previous versions (not packaged/published), the only possibility is VCS.

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/FileTable.js
alejandromumo: I got this snippet from `react-invenio-deposit` and, in fact, `file.checksum` never exists here. 

Therefore I am removing the checksum block

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
alejandromumo: `humanReadableBytes` returns a react fragment with the string inside..  Therefore can't be used as a string inside a string template.

It renders like:

```
Optional. Max attachments size: [object Object]
```

site/zenodo_rdm/__init__.py
zzacharo: I think that it started as an idea that you can publish this initially if the user wants but maybe is better to remove it and hardcode a version in the `setup.cfg`. @alejandromumo can you delete this file + the test? We will need to make the same changes in the `cookiecutter-invenio-rdm`

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/CategoryDropdown.js
zzacharo: this is the only way in react to render HTML...as long as we control this then is fine :)

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
alejandromumo: Managed to use it, however it required some changes in the backend (e.g. adding a error handler for `marshmallow.ValidationError`.

site/zenodo_rdm/__init__.py
ppanero: @zzacharo we can leave in cookiecutter, I do see that it can be useful... However, I'd remove it from this instance.

site/zenodo_rdm/support/support.py
alejandromumo: removed this one, the default error handler already does this.

site/zenodo_rdm/support/support.py
alejandromumo: it emulates the original one https://github.com/zenodo/zenodo/blob/master/zenodo/modules/support/views.py#L68, with more granular error catching (e.g. not sending the support email is not the same as not sending the confirmation email). 

 I agree that we lose the message and it will a nightmare to debug, especially since it is working with e-mail. Is there a standard way of logging exceptions in invenio?

site/tests/support/conftest.py
alejandromumo: it was a leftover, thanks!

site/tests/support/test_contact_form.py
alejandromumo: I agree, they were there as placeholders but it's better to simply create issues

site/zenodo_rdm/support/utils.py
alejandromumo: I don't think so, given that the form is currently validated by marshmallow. Adding commas yields an error in marshmallow so I will remove that `replace`.

site/tests/support/test_contact_form.py
alejandromumo: issue: https://github.com/zenodo/zenodo-rdm/issues/87

site/zenodo_rdm/support/support.py
alejandromumo: Thanks! I realized `500 errors` already have a default handler and we can work with that.

For `ValidationError` I added a custom handler. It returns an object simila to what we have in other modules, making the response serializable to formik. E.g. 

```json
{
  "errors": [
    {
      "field": "email",
      "messages": [
        "Not a valid email address."
      ]
    },
    {
      "field": "description",
      "messages": [
        "Description length must be bigger than 20 characters"
      ]
    }
  ]
}
```

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/CategoryDropdown.js
```diff
@@ -0,0 +1,76 @@
...
+class CategoryDropdown extends Component {
+
+    constructor(props) {
+        super(props);
+
+        let activeCategory;
+        const { categories, defaultCategory } = props;
+
+        if (!_isEmpty(categories)) {
+            activeCategory = categories.find((el) => (el.key === defaultCategory))
```
zzacharo, line 16: minor: You can use the [?](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) operator for simplifying the code maybe
```suggestion
            activeCategory = categories?.find((el) => (el.key === defaultCategory))
```

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/FileTable.js
```diff
@@ -0,0 +1,82 @@
...
+    decimalSizeDisplay: false
+}
+
+const FileTable = ({ filesList, handleDelete, decimalSizeDisplay }) => {
+    return (
+        <Table>
+            <Table.Body>
+                {filesList.map((file) => {
+                    return (
+                        <FileTableRow
```
zzacharo, line 58: minor: I would either split the `FileTableRow` in a new file or integrate the render method directly here.

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
```diff
@@ -0,0 +1,297 @@
...
+    withCredentials: true,
+    xsrfCookieName: "csrftoken",
+    xsrfHeaderName: "X-CSRFToken",
+    headers: {
+        "Accept": "application/json",
+        'Content-Type': 'multipart/form-data',
+    },
+};
+
+const http = axios.create(apiConfig);
```
zzacharo, line 28: major: We centralized [axios](https://github.com/inveniosoftware/react-invenio-forms/blob/master/src/lib/api/connector.js#L25) so you that one...You need only to modify the headers for your content type only!

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
```diff
@@ -0,0 +1,297 @@
...
+                            />
+                        }
+                    </div>
+                )
+            }}
+        </Dropzone>
+    );
+};
+
+FileUploader.propTypes = {
```
zzacharo, line 80: minor: maybe split it in another file

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
```diff
@@ -0,0 +1,297 @@
...
+
+            // TODO show error notification
+            console.error(errorMessage);
+        }
+    }
+
+    updateFiles = (files, formikSetFieldValue) => {
+        formikSetFieldValue("files", files)
+        this.setState({
+            files: files
```
zzacharo, line 143: why do we need to keep local state for files?

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
```diff
@@ -0,0 +1,297 @@
...
+                errorMessage = apiResponse.message || errorMessage;
+            }
+
+            // TODO show error notification
+            console.error(errorMessage);
+        }
+    }
+
+    updateFiles = (files, formikSetFieldValue) => {
+        formikSetFieldValue("files", files)
```
zzacharo, line 141: isn;t formik enough?

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
zzacharo: Try to have a component per file, at least when you have multiple components in the same file that are holding some state

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/FileTable.js
alejandromumo: Integrated the render method directly

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
alejandromumo: I was using it to render the list of currently accepted files.

However I realized that I can use `Formik.values` for that, therefore I am simplifying the component.

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
alejandromumo: indeed it is, thanks for the tip!

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
alejandromumo: Working on https://github.com/zenodo/zenodo-rdm/issues/84

PR #75, opened 2022-11-01
config: add custom fields

### LINKED ISSUE #53: zenodo: add custom metadata
https://github.com/zenodo/rdm-project/issues/1

slint:
* Closes #53.

### DIFF
--- a/invenio.cfg
+++ b/invenio.cfg
@@ -177,3 +177,108 @@ OAUTHCLIENT_AUTO_REDIRECT_TO_EXTERNAL_LOGIN = False  # autoredirect to external 
 USERPROFILES_READ_ONLY = False  # allow users to change profile info (name, email, etc...)
 
 THEME_SHOW_FRONTPAGE_INTRO_SECTION = False
+
+# Custom fields
+# -------------
+from invenio_records_resources.services.custom_fields import (
+    BaseCF,
+    DoubleCF,
+    ISODateStringCF,
+    KeywordCF,
+)
+from marshmallow import fields
+from marshmallow_utils.fields import SanitizedUnicode
+
+
+class RelationshipListCF(BaseListCF):
+    """Relationship list custom field."""
+
+    @property
+    def mapping(self):
+        return {
+            # NOTE: "object" wouldn't work for what we're querying
+            "type": "nested",
+            "properties": {
+                "subject": {"type": "keyword"},
+                "object": {"type": "keyword"},
+            },
+        }
+
+    @property
+    def field(self):
+        # For example see "Host of" on https://zenodo.org/record/3949282
+        return fields.List(fields.Nested({
+            'subject': fields.List(SanitizedUnicode(required=True)),
+            'object': fields.List(SanitizedUnicode(required=True)),
+        }))
+
+
+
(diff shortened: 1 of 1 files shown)


invenio.cfg
```diff
@@ -176,4 +176,104 @@ OAUTHCLIENT_AUTO_REDIRECT_TO_EXTERNAL_LOGIN = False  # autoredirect to external
...
+
+RDM_CUSTOM_FIELDS = [
+    # dwc
+    KeywordCF(name='dwc:basisOfRecord', multiple=True),
+    KeywordCF(name='dwc:catalogNumber', multiple=True),
+    KeywordCF(name='dwc:class', multiple=True),
+    KeywordCF(name='dwc:collectionCode', multiple=True),
+    KeywordCF(name='dwc:country', multiple=True),
+    KeywordCF(name='dwc:county', multiple=True),
+    KeywordCF(name='dwc:dateIdentified', multiple=True),
```
ppanero, line 230: minor: would it be useful to make them iso or edtf fields? that would allow for example to do range queries on them

invenio.cfg
```diff
@@ -176,4 +176,104 @@ OAUTHCLIENT_AUTO_REDIRECT_TO_EXTERNAL_LOGIN = False  # autoredirect to external
...
+    KeywordCF(name='dwc:decimalLatitude', multiple=True),
```
ppanero, line 231: minor: use `DoubleCF`?

invenio.cfg
```diff
@@ -176,4 +176,104 @@ OAUTHCLIENT_AUTO_REDIRECT_TO_EXTERNAL_LOGIN = False  # autoredirect to external
...
+    KeywordCF(name='dwc:scientificNameID', multiple=True),
+    KeywordCF(name='dwc:sex', multiple=True),
+    KeywordCF(name='dwc:specificEpithet', multiple=True),
+    KeywordCF(name='dwc:stateProvince', multiple=True),
+    KeywordCF(name='dwc:taxonID', multiple=True),
+    KeywordCF(name='dwc:taxonRank', multiple=True),
+    KeywordCF(name='dwc:taxonomicStatus', multiple=True),
+    KeywordCF(name='dwc:typeStatus', multiple=True),
+    KeywordCF(name='dwc:verbatimElevation', multiple=True),
+    KeywordCF(name='dwc:verbatimEventDate', multiple=True),
```
ppanero, line 261: minor: iso date?

ppanero (approved):
Minor suggestions on using more fields, though I guess they were already considered but not possible due to how the input data looks like.

Future: We should see if we can represent data better to allow e.g. doing queries like Zenodeo but in Zenodo.

invenio.cfg
slint: Agree. I'm only afraid because they're not validated at the moment, so we might shoot ourselves in the foot when doing the migration (but it's worth a try).

invenio.cfg
slint: Not for this one, because its `verbatim` (so could also be something like "1st of May, 1996"). There's also `dwc:eventDate` which I put as ISO date though.

Issue #70, opened 2022-10-27
frontpage: fix white color in footer sections


kpsherva:
![Image](https://user-images.githubusercontent.com/22594783/196763618-0ba9a210-b9b8-4ee6-8de4-9939161713a5.png)

vs zenodo sandbox

![Image](https://user-images.githubusercontent.com/22594783/196763770-1bc9c845-e06f-47fe-8231-caae5fbe7bd3.png)

kpsherva:
could not reproduce

PR #80, opened 2022-11-10
Footer: revert color white

Pineirin:
* closes: #70


kpsherva (approved):

kpsherva (commented):
after giving it a thought, it is not that simple. We want the text to be of color
@footerTextLightColor: #c9e5ff;
but not the section headers. Compare with the current production of zenodo

kpsherva (approved):

Issue #46, opened 2022-10-11
theming: deposit form


kpsherva:
# Preparation
- [x] make sure the layout is defined -  do we keep the RDM layout and styling?
- [x] gather all the additional information about changes/decisions regarding the elements of the frontpage (fill the `#decisions` section below)

# Elements
- [x] right side panel - access rights
- [x] fields (define order)

# Decisions

1. Hide access buttons to metadata (side panel)
2. hide metadata only button

For this task: Add config variables. Fix the backend part in a different task.

PR #90, opened 2022-11-15
assets: override metadata settings in the deposit form

jennur:
Closes https://github.com/zenodo/zenodo-rdm/issues/46

Overrides `<MetadataAccess />` and metadata-only toggle in `<FileUploaderToolbar />` in the deposit form.

**Needs**
- https://github.com/inveniosoftware/react-invenio-deposit/pull/602
- https://github.com/inveniosoftware/invenio-app-rdm/pull/1970

## Screenshot
Screenshot 2022-11-15 at 15 08 03

### DIFF
--- a/assets/js/invenio_app_rdm/deposit/override.js
+++ b/assets/js/invenio_app_rdm/deposit/override.js
@@ -1,0 +1,4 @@
+export const overriddenComponents = {
+    "ReactInvenioDeposit.MetadataAccess.layout": () => null,
+    "ReactInvenioDeposit.MetadataOnlyToggle.layout": () => null,
+}


invenio.cfg
```diff
@@ -283,3 +283,6 @@ RDM_CUSTOM_FIELDS = [
     # obo
     RelationshipListCF(name='obo:RO_0002453'),
 ]
+
+APP_RDM_DEPOSIT_REQUIRE_FILES = True
+APP_RDM_DEPOSIT_FULL_RECORD_VISIBILITY = False
```
kpsherva, line 288: is there any reason why we add config variables instead of using overridable ?

kpsherva: After conversation with @slint : we agreed to use Overridable.parametrize function to overwrite the components' props instead of adding a config variable, since config variables should be reserved for full features, not only disabling the frontend components. The solution proposed in these PRs does not provide the backend validation, only hides the access field from the user (while it is still possible to bypass in REST API) there should not be any flag indicating it is a full feature, untill we provide a proper implementation, which includes records permission policy.
Actionable now: change the implementation to use overridable's parametrise function.
The overriden component code should be added in Zenodo's equivalent of https://github.com/inveniosoftware/cookiecutter-invenio-rdm/tree/master/%7B%7Bcookiecutter.project_shortname%7D%7D/site/%7B%7Bcookiecutter.package_name%7D%7D/assets/semantic-ui/js/%7B%7Bcookiecutter.package_name%7D%7D

kpsherva (approved):

PR #95, opened 2022-11-16
refactor/consolidation of code

ppanero:
run `isort --profile black .` to fix imports


migration/transform/base.py
```diff
@@ -0,0 +1,48 @@
...
+        """Transform the id."""
+        pass
+
+    @abstractmethod
+    def _record(self, entry):
+        """Transform the record."""
+        pass
+
+    @abstractmethod
+    def _draft(self, entry):
```
ppanero, line 26: not 100% convinced, this assumes that from one entry we can extract all that information. When in reality this comes because we are transforming Zenodo records (which have `_deposit`). However, it makes sense to have record, draft, parent and files as a unit...

migration/load/dataclasses.py
```diff
@@ -0,0 +1,105 @@
+"""Dataclasses for generating table rows."""
+
+from dataclasses import InitVar, dataclass
```
ppanero, line 3: naming clash with the Python package... How can we call this? _models_?

migration/load/postgresql/tables/records.py
```diff
@@ -0,0 +1,211 @@
...
+        "obj_type": "rec",
+        "pid_type": "recid",
+        "status": "R",
+    }
+
+def _generate_uuid(data):
+    return str(uuid.uuid4())
+
+
+class RDMRecordTableLoad(PostgreSQLTableLoad):
```
zzacharo, line 34: Isn;t this the concept of a "stream"?

migration/load/postgresql/postgresql.py
```diff
@@ -0,0 +1,149 @@
...
+                    'latest_index': 'record_index',
+                    'latest_id': 'record id',
+                }
+        }
+        """
+        self.cur_dir = Path(__file__).parent
+        self.output_dir = self.cur_dir / f"tables{_ts(iso=False)}"
+
+        # tables to prepare from loaders
+        self._tables = tables or [
```
zzacharo, line 48: this should be a "stream" no?

migration/load/postgresql/tables/records.py
ppanero: Not really, the stream will be the bigger picture part. This could be a "substream". Once the code is __fully__ done, this will hopefully be more clear. Same applied to comment below.

PR #97, opened 2022-11-18
github: removed confirmation required for login

### LINKED ISSUE #78: configure github and orcid login

alejandromumo:
closes #78 

Notes:

- The config `SECURITY_LOGIN_WITHOUT_CONFIRMATION` forces the user to confirm its account before logging in. For Github, that flow is not what an user would expect. After logging in with Github, the user should be automatically logged in.
- Even though this config affects other login methods, this config is currently set to `True` in other production instances as well.
- The "real" fix here would be for the github `signup_handler.info` [handler](https://github.com/inveniosoftware/invenio-oauthclient/blob/master/invenio_oauthclient/contrib/github.py#L164) to automatically set the user as confirmed. However that does not work right now. Since this would be a core change, I created an [issue](https://github.com/inveniosoftware/invenio-oauthclient/issues/286) to analyse this separately.


kpsherva:
what about ORCid Login? does it work?

ppanero (approved):
notes:
- users will still need to confirm account before certain actions like create a community
- users will be able to login (otherwise the post-registration is not user friendly)
- orcid login is not working (same as in Zenodo sandbox, waiting for ORCID people to reply)

slint (approved):

PR #98, opened 2022-11-18
support: added recaptcha

### LINKED ISSUE #83: support form: add re-captcha
The new support form does not have a re-captcha yet implemented.

The new support form is currently implemented in https://github.com/zenodo/zenodo-rdm/pull/61 

In this issue:

- [x] Evaluate how re-captcha is currently implemented
- [x] Migrate the re-captcha to the new form

See the current re-captcha in action: https://sandbox.zenodo.org/support
See the current implementation in: 
- https://github.com/zenodo/zenodo/blob/master/zenodo/modules/support/templates/zenodo_support/contact_form.html

alejandromumo:
closes #83

### DIFF
--- a/invenio.cfg
+++ b/invenio.cfg
@@ -51,6 +51,16 @@ APP_DEFAULT_SECURE_HEADERS = {
             "blob:",            # for pdf preview
             # Add your own policies here (e.g. analytics)
         ],
+        "frame-src": [
+            "'self'",
+            "https://www.google.com/recaptcha/",
+            "https://recaptcha.google.com/recaptcha/",
+        ],
+        "script-src": [
+            "'self'",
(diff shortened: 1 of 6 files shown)


alejandromumo:
The google  `client key` is being injected in the DOM, leaving it open to read. That is not ideal in terms of security. 

I tried to pass the key using an environment variable but I could not make it work. It seems [invenio-assets](https://github.com/inveniosoftware/invenio-assets/blob/master/invenio_assets/assets/build/webpack.config.js#L181) is currently replacing `process.env` for `process.env.NODE_ENV` and I don't find any other way to pass other variables.

The PR is therefore in `draft` until this can be properly implemented (without exposing the client key)

PR #100, opened 2022-11-21
Data migration zenodo

ppanero:
zenodo specifics for migration:
- requires https://github.com/inveniosoftware/invenio-rdm-migrator/pull/1 made into a proper module and installable
- requires https://github.com/inveniosoftware/invenio-rdm-migrator/pull/23
- requires https://github.com/inveniosoftware/invenio-rdm-migrator/pull/27

### DIFF
--- a/site/zenodo_rdm/migrator/__init__.py
+++ b/site/zenodo_rdm/migrator/__init__.py
@@ -1,0 +1,15 @@
+import sys
+
+from stream import RecordStream
+
+
+if __name__ == "__main__":
+
+    filename = sys.argv[1]
+    cleanup = len(sys.argv) > 2  # if there is something we assume is True/--cleanup
+    stream = RecordStream(
+        filename=filename,
+        db_uri="postgresql://zenodo:zenodo@localhost:5432/zenodo",
+        output_path="/Users/ppanero/Workspace/zenodo/zenodo-rdm/migration/data/",
+    )    
+    stream.run(cleanup=cleanup)
--- a/site/zenodo_rdm/migrator/__init__.py
+++ b/site/zenodo_rdm/migrator/__init__.py
@@ -1,15 +1,14 @@
 import sys
 
-from stream import RecordStream
+from invenio_rdm_migrator.streams import Runner
+
+from stream import RecordStreamDefinition
 
 
 if __name__ == "__main__":
 
-    filename = sys.argv[1]
-    cleanup = len(sys.argv) > 2  # if there is something we assume is True/--cleanup
-    stream = RecordStream(
-        filename=filename,
-        db_uri="postgresql://zenodo:zenodo@localhost:5432/zenodo",
-        output_path="/Users/ppanero/Workspace/zenodo/zenodo-rdm/migration/data/",
+    runner = Runner(
+        stream_definitions=[RecordStreamDefinition],
+        config_filepath=sys.argv[1],
     )    
-    stream.run(cleanup=cleanup)
+    runner.run()
--- a/site/zenodo_rdm/migrator/extract.py
+++ b/site/zenodo_rdm/migrator/extract.py
@@ -1,0 +1,19 @@
+import json
+
+from invenio_rdm_migrator.extract import Extract
+
+
+class JSONLExtract(Extract):
+    """Data extraction from JSONL files."""
+
+    def __init__(self, filepath):
+        """Constructor."""
+        self.filepath = filepath
+
+    def run(self):
+        """Yield one element at a time."""
+        with open(self.filepath, 'r') as reader:
+            for l in reader:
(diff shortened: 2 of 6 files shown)


site/zenodo_rdm/migrator/extract.py
```diff
@@ -0,0 +1,19 @@
+import json
+
+from invenio_rdm_migrator.extract import Extract
+
+
+class JSONLExtract(Extract):
```
slint, line 6: minor: could this be part of `invenio_rdm_migrator`?

migration/requirements.txt
```diff
@@ -0,0 +1,5 @@
+nameparser
```
slint, line 1: minor: This is probably the only dependency that has to be moved to the `Pipfile`. `psycopg` should be part of `invenio_rdm_migrator`

translations/de/LC_MESSAGES/messages.po
```diff
@@ -0,0 +1,169 @@
+# German translations for PROJECT.
```
slint, line 1: minor: not sure where this came from but looks a bit weird/generic.

translations/de/LC_MESSAGES/messages.po
```diff
@@ -0,0 +1,169 @@
...
+"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
+"Language: de\n"
+"Language-Team: de <LL@li.org>\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=utf-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Generated-By: Babel 2.11.0\n"
+
+#: /Users/ppanero/Workspace/zenodo/zenodo-rdm/invenio.cfg:88
```
slint, line 21: minor: Also has some original path-specific things which probably shouldn't be here

site/zenodo_rdm/migrator/extract.py
ppanero: I decided not to since it is slightly custom due to our dump. We are not using jsonlines but doing the custom replace and then loading the JSON. Therefore, it would only work with a specific type of `""` input.

translations/de/LC_MESSAGES/messages.po
ppanero: 👀 no idea how this got here, should not indeed

site/zenodo_rdm/migrator/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+from stream import RecordStream
+
+
+if __name__ == "__main__":
+
+    filename = sys.argv[1]
+    cleanup = len(sys.argv) > 2  # if there is something we assume is True/--cleanup
+    stream = RecordStream(
+        filename=filename,
+        db_uri="postgresql://zenodo:zenodo@localhost:5432/zenodo",
```
ppanero, line 12: major: both are hardcoded to my current setup. How do we want to run this? click? or for now I can add `argv[2]` and `argv[3]`?

slint: Discussed IRL:

- we'll keep the `__main__` + `argv` usage in the instances for now
- we configure the classes for each stream (users, communities, records) in Python in the instance
- we introduce a YAML config to pass specific parameters to the ETL classes (e.g. `db_uri`, `filename` for the JSONL extract, `validate=True`, `cleanup=True`, etc.)

site/zenodo_rdm/migrator/stream.py
```diff
@@ -0,0 +1,17 @@
...
+
+class RecordStream(Stream):
+    """ETL stream for Zenodo to RDM records."""
+
+    def __init__(self, filename, db_uri, output_path):
+        """Constructor."""
+        super().__init__(
+            extract=JSONLExtract(filename),
+            transform=ZenodoToRDMRecordTransform(),
+            load=RDMRecordCopyLoad(db_uri, output_path),
```
kpsherva, line 17: IMO it is input_path since its the input to load from no?

ppanero: As any other question in the migration sprint, the answer is "yes and no". While you are correct on the fact that we are reading/loading data from that directory, namely the csv files to `COPY`; we are also writing/dumping files there, namely those csv files produced by the `prepare` step.

So yes, we are loading, but we are also dumping. What about `tmp_dir` or `data_dir` (dir or path)?

ppanero: Fixed, discussed IRL it's `tmp_dir`

ppanero:
included in https://github.com/zenodo/zenodo-rdm/pull/107

PR #107, opened 2022-11-24
migrator: users stream

slint:
Closes https://github.com/inveniosoftware/invenio-rdm-migrator/issues/2.

- migrator: recrods stream specifics
- migrator: use streams runner and yaml config file
- migration: add users extract and transform


ppanero (approved):
Removed the kwargs passing, to enforce them. It worked for me, if for your it fails we should investigate a bit... I have the changes saved in a local branch in case of need.

Runs:
```
Records - Execution time: 0:00:31.087643
Users - Execution time: 0:00:05.079141
```

Only found an issue with an existing user id, we shouldn't have that in RDM since db should be clean.

PR #108, opened 2022-11-24
add custom oai sets

anikachurilova:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1931


slint (approved):
Thanks @anikachurilova! Updated the queries and added some additional sets from Zenodo legacy.

PR #110, opened 2022-11-25
docs: how to carry out a test migration to paas instance

ppanero:
- requires #109

### DIFF
--- a/site/zenodo_rdm/migrator/transform.py
+++ b/site/zenodo_rdm/migrator/transform.py
@@ -202,7 +202,7 @@ class ZenodoUserEntry(UserEntry):
 
     def _version_id(self, entry):
         """Returns the version id."""
-        return entry.get("version_id", 0)
+        return entry.get("version_id", 1)
 
     def _email(self, entry):
         """Returns the email."""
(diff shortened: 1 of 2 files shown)


site/zenodo_rdm/migrator/transform.py
```diff
@@ -237,8 +237,8 @@ def _profile(self, entry):
     def _preferences(self, entry):
         """Returns the preferences."""
         return {
-            "visibility": "private",
-            "email_visibility": "private",
+            "visibility": "restricted",
```
alejandromumo, line 240: Small comment: it was recently discussed whether the default's profile visibility should be configured per instance (e.g. a config). 

https://github.com/inveniosoftware/invenio-app-rdm/issues/1779

Not blocking, just leaving a notice for later.

alejandromumo (approved):

kpsherva (approved):

site/zenodo_rdm/migrator/transform.py
slint: Transform will be hardcoded per instance, and since we're bypassing the service/schemas config wouldn't affect this class.

The plan is in general to leave the transform/extract logic free from any application logic, so that it can be fast/simple and focus on data mangling.

Validation (of different levels/types) will anyways take place on the Load step.

PR #119, opened 2022-12-01
sidebar segment: correct placeholder bg color

jennur:
The placeholder background color is currently white, which does not look great on the gray bg for the sidebar segments in Zenodo. 

This PR aligns the bg color of the placeholder with the sidebar segment.

## Screenshot
Screenshot 2022-12-01 at 10 31 59

### DIFF
--- a/assets/less/zenodo-rdm/elements/segment.overrides
+++ b/assets/less/zenodo-rdm/elements/segment.overrides
@@ -8,8 +8,8 @@
 .ui.segment.rdm-sidebar {
   background-color: @rdmSidebarSegmentColor;
   border: 1px solid @rdmSidebarSegmentBorderColor;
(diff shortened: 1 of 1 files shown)


assets/less/zenodo-rdm/elements/segment.overrides
```diff
@@ -28,16 +28,24 @@
       padding: 0;
     }
   }
+
+  .ui.placeholder {
+    &,
```
kpsherva, line 33: & will resolve to the parent, so you could use the parent directly. Unless it is a syntax sugar :)

jennur: The thing is that I want it also to apply to the parent, without defining it twice (on the parent, and on the children selectors)

zzacharo (approved):

Issue #111, opened 2022-11-28
prepare sandbox for zenodo presentation


zzacharo:
- [ ] add a banner indicating that it is a demo website
- [x] dump 0.5M - 1M records belonging to Biodiversity and NASA communities
- [x] load records to Zenodo RDM sandbox

kpsherva:
first task moved to another issue

PR #125, opened 2022-12-01
Add communities to records

zzacharo:
addresses #111


site/zenodo_rdm/migrator/README.md
```diff
@@ -65,6 +70,18 @@ invenio vocabularies import -v names -f ./app_data/vocabularies-future.yaml  # z
 python -m zenodo_rdm.migrator /path/to/streams.yaml
 ```
 
+```bash
+# Manual COPY ...
+export DB_URI=${SQLALCHEMY_DATABASE_URI}
```
ppanero, line 75: curiosity: this is not done programmatically because?

site/zenodo_rdm/migrator/dump_communities.py
```diff
@@ -0,0 +1,11 @@
+import yaml
+from pathlib import Path
+from invenio_communities.communities.records.models import CommunityMetadata
+community_map = {comm.slug: str(comm.id) for comm in CommunityMetadata.query.all()}
+streams_path = str(Path('site/zenodo_rdm/migrator/streams.yaml').absolute())
+streams = {}
+with open(streams_path, 'r') as fp:
```
ppanero, line 7: I guess there is no way to overwrite only one key, that's why read-then-write?

ppanero (approved):
why is the empty `.wipe_recreate.sh` file needed?

site/zenodo_rdm/migrator/dump_communities.py
zzacharo: didn't find one, but still this is a temp solution for now....

site/zenodo_rdm/migrator/README.md
zzacharo: this is needed for the manual copy of the csv via the psql only

Issue #122, opened 2022-12-01
ui: record landing page details sidebar titles font style


slint:
I think these should be bold/strong or similar:

ZenodoRDM:

![Image](https://user-images.githubusercontent.com/1814661/205074083-02148f34-4158-473a-8a35-d70994914c0c.png)

vs. 

InvenioRDM:

![Image](https://user-images.githubusercontent.com/1814661/205074244-9c5c704a-c703-4bd9-8a33-580a6de5fd51.png)

Pineirin:
Couldn't reproduce.
![Screenshot from 2022-12-02 17-35-51](https://user-images.githubusercontent.com/25476209/205341069-ec63f12c-bb18-4771-82be-c95da6bccd96.png)

PR #130, opened 2022-12-05
Record details: make headers bold

Pineirin:
closes: #122
Screenshot from 2022-12-05 17-14-04


kpsherva:
could you add a screenshot?

assets/less/zenodo-rdm/views/item.overrides
```diff
@@ -0,0 +1,9 @@
+/***********************************************
+         Zenodo RDM Item Overrides
+***********************************************/
+
+#record-details {
+    .header {
```
zzacharo, line 6: that changes every header in the record details page. Is this what we want? Can you add a screenshot with the whole page and compare it with zenodo? I would be tempted to change only the header in the sidebar as the issue was stating...

Pineirin: Screenshot from 2022-12-06 13-45-01

Unless we are duplicating ids. The records-details id only makes reference to the section we want to change, and not to the whole page.

kpsherva: checked, the `#record-details` is referring only to that section.
however I think this is too broad, since semantic ui classes are usually marked as `.ui.header`.
in addition the file misses formatting (missing semi-colon)

kpsherva (approved):

PR #131, opened 2022-12-05
frontpage: make recent uploads query configurable

zzacharo:


invenio.cfg
```diff
@@ -290,3 +290,7 @@ RDM_CUSTOM_FIELDS = [
     # obo
     RelationshipListCF(name='obo:RO_0002453'),
 ]
+
+# Search query of recent uploads
+# Defaults to newest records search
+FRONTAGE_RECENT_UPLOADS_QUERY = "/api/records?sort=newest&size=10"
```
kpsherva, line 296: could we place it in the template directly, instead of adding the config var?

zzacharo: nope because we want to retrieve a specific community's records so we avoid showing spams. The url is including the id of the community so we will need to set it via environment variable!

kpsherva: 👍 in this case I suggest to prefix the name of the variable with `ZENODO_` so that we know it is coming from the instance, not any of the modules

invenio.cfg
```diff
@@ -290,3 +290,7 @@ RDM_CUSTOM_FIELDS = [
     # obo
     RelationshipListCF(name='obo:RO_0002453'),
 ]
+
+# Search query of recent uploads
+# Defaults to newest records search
+ZENODO_FRONTAGE_RECENT_UPLOADS_QUERY = "/api/records?sort=newest&size=10"
```
kpsherva, line 296:
```suggestion
ZENODO_FRONTPAGE_RECENT_UPLOADS_QUERY = "/api/records?sort=newest&size=10"
```

Issue #135, opened 2022-12-05
citation-box: styling misalignments


zzacharo:
![Screenshot 2022-12-05 at 17 33 43](https://user-images.githubusercontent.com/22594783/205694790-47bf067b-d6dc-4bef-aa09-866cef386551.png)

Compare with zenodo prod on the right side!

PR #138, opened 2022-12-06
landing page: add citations search box

jennur:
Closes https://github.com/zenodo/zenodo-rdm/issues/135

## Screenshot
Screenshot 2022-12-06 at 11 26 47


kpsherva:
looks good!, only a small comment, see: https://github.com/zenodo/zenodo-rdm/issues/135#issuecomment-1339268343

assets/less/zenodo-rdm/globals/site.overrides
```diff
@@ -106,3 +106,6 @@ a.inverted {
   height: 0.85em;
   width: 1.2em;
 }
+.uppercase {
+  text-transform: uppercase !important;
```
kpsherva, line 110: invenio-theme candidate?

zzacharo: I wouldn't put it if that's only for this case as I think is a hack to uppercase the values...but indeed seems like a generic class to add in invenio-theme!

jennur:
## Update
- Had to move the accordion to React to get the counter in at the panel title.

The accordion from basic (html, jQuery) SUI vs React SUI behaves differently, as the React one apparently does not have a transition option like the basic one does. If it's not super important, (for now) maybe we can leave it as is? I'm not sure how to solve this issue nicely atm, so please lmk if you have suggestions.

Screenshot 2022-12-06 at 15 22 25

PR #141, opened 2022-12-07
cli: add script to delete a record

zzacharo:


ppanero:
Is this intended for cleanup? there is a cleanup flag in invenio-rdm-migrator. Things could be deleted afterawrds, so maybe we want to hook the interface with this?

PR #154, opened 2022-12-14
search: added legacy search map

### LINKED ISSUE #102: migration: search redirections
If we compare the [Zenodo search guide](https://help.zenodo.org/guides/search/) and the one that [comes with InvenioRDM](https://inveniordm.web.cern.ch/help/search), there are quite some changes in terms of nested fields (`title` vs `metadata.title`) and semantically equivalent fields (`keyword` vs `subjects`).

Many of these changes are related to the different search index mappings, but also migrated/changed fields. We have to:

- [x] Identify the differences between mappings
- [ ] Make a decision to either...
  - [ ] Use the query parser to translate the fields (see [users example](https://github.com/inveniosoftware/invenio-users-resources/blob/ae6b45e548b554b737b83c6109de4ff4fb14a04a/invenio_users_resources/services/users/config.py#L39-L52))
  - [ ] Use `copy_to` in the mappings to create fields "aliases"; not ideal since it means overriding mappings (which I'm not sure if it's possible atm)

alejandromumo:
closes #102

### DIFF
--- a/invenio.cfg
+++ b/invenio.cfg
@@ -15,6 +15,10 @@ by installer are included here.
 from flask_babelex import lazy_gettext as _
 from invenio_oauthclient.contrib.orcid import ORCIDOAuthSettingsHelper
 from invenio_oauthclient.contrib import github, openaire_aai
+from invenio_records_resources.services.records.queryparser import (
+    QueryParser,
+    SearchFieldTransformer,
+)
 
 # Flask
 # =====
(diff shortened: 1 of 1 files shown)


alejandromumo:
TODO: 

- Some maps do not have an equivalent term.
- Some queries do not work (e.g. `metadata.contributors.affiliations.name`)

zzacharo (approved):

zzacharo:
Address comment https://github.com/zenodo/zenodo-rdm/issues/102#issuecomment-1352886639

zzacharo (approved):

Issue #153, opened 2022-12-14
email templates are not correctly displayer


ppanero:
<img width="1753" alt="Screenshot 2022-12-14 at 11 09 00" src="https://user-images.githubusercontent.com/6756943/207567123-1040095f-a124-4646-93dd-1c5a33fc44ca.png">

ppanero:
closing this in favour of https://github.com/zenodo/zenodo-rdm/issues/137, since it has more details

Issue #137, opened 2022-12-06
Improve account creation related email templates


anikachurilova:
- make templates compatible with different screen sizes
- check compatibility with different email clients (Gmail, Outlook, etc). Ex: sending emails to [Mailtrap](https://mailtrap.io/home) can help checking HTML-elements on different email clients:
<img width="1670" alt="Screenshot 2022-12-06 at 09 57 05" src="https://user-images.githubusercontent.com/61321254/205868106-12b3e806-e418-4c96-89bd-c06b03014d97.png">

<img width="978" alt="Screenshot 2022-12-06 at 10 02 09" src="https://user-images.githubusercontent.com/61321254/205868141-ce29a328-fe73-4785-a0e4-84841a7a48be.png">

PR #169, opened 2023-01-10
email templates: improve compatibility with different clients

jennur:
Closes https://github.com/zenodo/zenodo-rdm/issues/153
Closes https://github.com/zenodo/zenodo-rdm/issues/137

- Fixes button position bug (`<a>` was inline by default, needed to be `inline-block` because of padding).
- Moves styling from `<style>`-block to inline styling because of lack of support for the style-block on some e-mail clients.
- Changes logo to png-version because of compatibility issues on some clients with svg files.
- Changes width from 50% to max 500px to make it responsive.

Used [Mailtrap](https://mailtrap.io/) and [MailHog](https://github.com/mailhog/MailHog) for testing.
## Screenshots

Screenshot 2023-01-10 at 15 17 27
Screenshot 2023-01-10 at 15 17 00
Screenshot 2023-01-10 at 15 14 40
Screenshot 2023-01-10 at 15 01 46
Screenshot 2023-01-10 at 15 01 41
Screenshot 2023-01-10 at 15 01 34


ppanero (commented):
Nice job! Two tiny things:
- Would it make sense to leave the two images? svg and png? wondering if we would e.g. use the svg for the frontpage?
- There is a `.DS_Store` file. Could you remove it, and maybe update the gitignore to avoid them 😅

jennur:
> Nice job! Two tiny things:
> 
> * Would it make sense to leave the two images? svg and png? wondering if we would e.g. use the svg for the frontpage?
> * There is a `.DS_Store` file. Could you remove it, and maybe update the gitignore to avoid them 😅

Wopsie, thanks!! I didn't notice. For the svg file, I didn't remove it, I just added the png :)

kpsherva (approved):

PR #177, opened 2023-01-17
fixtures: fix communities YAML

slint:


zzacharo (approved):
I don;t understand the change from the diff 😅 How did you find it? Maybe I miss some VSCode configuration for the yaml plugin....

slint:
There are double-quotes in one of the YAML text fields, so we have to escape. A simplified diff would be:

```diff
-body: "The phrase "carry on" is commonly used"
+body: "The phrase \"carry on\" is commonly used"
```

PR #196, opened 2023-02-23
contrib: added journal custom field.

alejandromumo:
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1191

- Added `Journal` custom field to zenodo RDM

### DIFF
--- a/invenio.cfg
+++ b/invenio.cfg
@@ -20,6 +20,14 @@ from invenio_records_resources.services.records.queryparser import (
     SearchFieldTransformer,
 )
 
+# Journal CF
+
+from invenio_rdm_records.contrib.journal.custom_fields import (
+    JOURNAL_NAMESPACE,
+    JOURNAL_CUSTOM_FIELDS,
+    JOURNAL_CUSTOM_FIELDS_UI
+)
+
 # Flask
 # =====
 # See https://flask.palletsprojects.com/en/1.1.x/config/
@@ -162,6 +170,13 @@ ZENODO_LEGACY_SEARCH_MAP = {
     "title": "metadata.title",
     "type": "metadata.resource_type.props.type",
     "version": "versions.index",
+    # journal custom field mappings
+    "journal.title": "custom_fields.journal\:journal.title",
+    "journal.pages": "custom_fields.journal\:journal.pages",
+    "journal.volume": "custom_fields.journal\:journal.volume",
+    "journal.year": "custom_fields.journal\:journal.year",
+    "journal.issn": "custom_fields.journal\:journal.issn",
+
 }
 
 RDM_SEARCH = {
@@ -276,6 +291,8 @@ RDM_NAMESPACES = {
     "openbiodiv": "http://openbiodiv.net/ontology#",
     "obo": "http://purl.obolibrary.org/obo/",
     "dc": "http://purl.org/dc/terms/",
+    # journal cf
+    **JOURNAL_NAMESPACE,
 }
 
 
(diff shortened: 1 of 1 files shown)


invenio.cfg
```diff
@@ -341,3 +341,26 @@ RDM_CUSTOM_FIELDS = [
...
+]
+
+RDM_CUSTOM_FIELDS = [
+    *RDM_CUSTOM_FIELDS,
+    *JOURNAL_CUSTOM_FIELDS,
+]
+
+RDM_NAMESPACES = {
+    **RDM_NAMESPACES,
+    **JOURNAL_NAMESPACE,
```
zzacharo, line 365: why don't you add it directly when we define the `RDM_NAMESPACES`?

invenio.cfg
```diff
@@ -341,3 +341,26 @@ RDM_CUSTOM_FIELDS = [
...
+    JOURNAL_CUSTOM_FIELDS_UI
+)
+
+RDM_CUSTOM_FIELDS_UI = [
+    JOURNAL_CUSTOM_FIELDS_UI
+]
+
+RDM_CUSTOM_FIELDS = [
+    *RDM_CUSTOM_FIELDS,
+    *JOURNAL_CUSTOM_FIELDS,
```
zzacharo, line 360: Same as below

invenio.cfg
```diff
@@ -341,3 +341,26 @@ RDM_CUSTOM_FIELDS = [
...
+# Journal CF
+
+from invenio_rdm_records.contrib.journal.custom_fields import (
+    JOURNAL_NAMESPACE,
+    JOURNAL_CUSTOM_FIELDS,
+    JOURNAL_CUSTOM_FIELDS_UI
+)
+
+RDM_CUSTOM_FIELDS_UI = [
+    JOURNAL_CUSTOM_FIELDS_UI
```
zzacharo, line 355: is this an object or a list?

invenio.cfg
```diff
@@ -341,3 +341,26 @@ RDM_CUSTOM_FIELDS = [
 # Search query of recent uploads
 # Defaults to newest records search
 ZENODO_FRONTPAGE_RECENT_UPLOADS_QUERY = "/api/records?sort=newest&size=10"
+
+
+# Journal CF
+
+from invenio_rdm_records.contrib.journal.custom_fields import (
```
zzacharo, line 348: move the imports in the beginning of the file?

zzacharo (commented):
@slint probably out of scope but the custom fields are getting out of hand I think in invenio.cfg....maybe we should move them to `site` folder.....

invenio.cfg
alejandromumo: `JOURNAL_CUSTOM_FIELDS_UI` is an object, while `RDM_CUSTOM_FIELDS_UI` is a [list](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/config.py#L405)

invenio.cfg
alejandromumo: Nice catch, it was a leftover from testing codemeta + journal at the same time.
I will move it!

invenio.cfg
alejandromumo: Nice catch, it was a leftover from testing codemeta + journal at the same time.
I will move it!

ppanero (approved):
@alejandromumo could you also note down that we might want to move this to site folder? maybe worth asking for a discussion time and also add the naming question.

PR #198, opened 2023-02-23
contrib: added meeting custom field

alejandromumo:
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1211

**REQUIRES**

- https://github.com/inveniosoftware/react-invenio-forms/pull/192
- https://github.com/inveniosoftware/invenio-app-rdm/pull/2102

### DIFF
--- a/invenio.cfg
+++ b/invenio.cfg
@@ -21,11 +21,17 @@ from invenio_records_resources.services.records.queryparser import (
 )
 
 # Journal CF
-
 from invenio_rdm_records.contrib.journal.custom_fields import (
     JOURNAL_NAMESPACE,
     JOURNAL_CUSTOM_FIELDS,
     JOURNAL_CUSTOM_FIELDS_UI
+)
+
+# MEETING CF
+from invenio_rdm_records.contrib.meeting.custom_fields import (
+    MEETING_NAMESPACE,
+    MEETING_CUSTOM_FIELDS,
+    MEETING_CUSTOM_FIELDS_UI
 )
 
 # Flask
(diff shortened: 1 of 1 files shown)


kpsherva (approved):
LGTM! However I prefer to have also a second opinion

invenio.cfg
```diff
@@ -336,8 +353,16 @@ RDM_CUSTOM_FIELDS = [
     KeywordCF(name='openbiodiv:TaxonomicConceptLabel', multiple=True),
     # obo
     RelationshipListCF(name='obo:RO_0002453'),
+    # meeting
+    *MEETING_CUSTOM_FIELDS,
 ]
 
+# Custom fields UI components
+RDM_CUSTOM_FIELDS_UI = [
+    MEETING_CUSTOM_FIELDS_UI
```
zzacharo, line 362: Will that conflict with the journal config PR?

alejandromumo: Yes, I will rebase it though when things are merged.

ppanero, line 362: Should we make consistent use of lists. In this case meeting only has one section, however, what if they were more? then we end up with a mix of `*` and non `*`. I see two options:
a) Enforce one section per `XYZ_CUSTOM_FIELDS_UI` in which case i'd rename to `MEETING_CUSTOM_FIELDS_UI_SECTION`
b) Make `MEETING_CUSTOM_FIELDS_UI` a list then `*` here.

alejandromumo: I see your point and I do agree, right now we can only have one section per custom field. However, and correct me if I am wrong, that behavior is part of the custom fields interface (e.g. `XYZ_CUSTOM_FIELDS_UI` is an object with  a property `section` and `fields`). 

For `RDM_CUSTOM_FIELDS_UI`, it will look like this:

```python
RDM_CUSTOM_FIELDS_UI = [
    MEETING_CUSTOM_FIELDS_UI,
    JOURNAL_CUSTOM_FIELDS_UI,
    ...
]
```

ppanero: My question is not on any inner behaviours, it focuses on naming/user experience. If I have:
- RDM_CUSTOM_FIELDS_UI
- MEETING_CUSTOM_FIELDS_UI
- JOURNAL_CUSTOM_FIELDS_UI
- XYZ_CUSTOM_FIELDS_UI

It gives me the impresion that they are analog, meaning they can be replaced. For example:

```python
RDM_CUSTOM_FIELDS_UI  = XYZ_CUSTOM_FIELDS_UI
```

When in reality is not true, because the last 3 of the list above are "parts" of the first one. That's why I wanted to suffix with `SECTION`. However, the name becomes long.

PR #202, opened 2023-02-27
contrib: added imprint custom field

alejandromumo:
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1215

### DIFF
--- a/invenio.cfg
+++ b/invenio.cfg
@@ -20,6 +20,13 @@ from invenio_records_resources.services.records.queryparser import (
     SearchFieldTransformer,
 )
 
+# Imprint CF
+from invenio_rdm_records.contrib.imprint.custom_fields import (
+    IMPRINT_NAMESPACE,
+    IMPRINT_CUSTOM_FIELDS,
+    IMPRINT_CUSTOM_FIELDS_UI
+)
+
 # Journal CF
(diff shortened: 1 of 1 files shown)


invenio.cfg
```diff
@@ -162,6 +169,12 @@ ZENODO_LEGACY_SEARCH_MAP = {
     "title": "metadata.title",
     "type": "metadata.resource_type.props.type",
     "version": "versions.index",
+    "part_of.title": "custom_fields.imprint\:imprint.title",
```
ppanero, line 172: question: why `part_of`?

ppanero (approved):
just confirm that is `part_of` and not `imprint`, otherwise all good!

invenio.cfg
alejandromumo: That's how this field was implemented in zenodo.

See the [search guide](https://help.zenodo.org/guides/search/) or the [jsonschema](https://github.com/zenodo/zenodo/blob/d910dbc0c29f9829ca24a2239f4f01cf295c9687/zenodo/modules/records/jsonschemas/records/record-v1.0.0.json#L848) .

Even though in the UI all these fields (`imprint` and `part_of`) are together, in the backend they were split back in Zenodo. After discussing IRL, we decided to merge them under a single custom field `imprint` and guarantee retro-compatibility for the search.

PR #220, opened 2023-03-14
data: add biohackathon community

ppanero:
required for latest migration dump


ppanero:
not needed, migrator logic temporarily changed to only used fixtured communities. Next migration will have multiple communities thus communities will also be migrated.

PR #222, opened 2023-03-15
config: add sorting for conference, journal, versions and publication…

ppanero:
- closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2030

**CURRENT STATUS**
Screenshot 2023-03-24 at 14 49 02

**IDEAL**

We can treat it as a separate issue since it is purely UX/UI, extending the sort widget width to 16em (instead of 14) would fix it.
Screenshot 2023-03-24 at 14 51 21

**Open questions**
- "Newest" / "Oldest" does not really make sense in this cases since they are sorting over numbers/strings. Should the be "Descending" or "Desc"?
- Do we want to test with icons? (this would require changes in other repositories/sortings too)

### DIFF
--- a/invenio.cfg
+++ b/invenio.cfg
@@ -16,6 +16,9 @@ from invenio_i18n import lazy_gettext as _
 from invenio_oauthclient.contrib.openaire_aai import REMOTE_SANDBOX_APP
 from invenio_oauthclient.contrib.orcid import ORCIDOAuthSettingsHelper
 from invenio_oauthclient.contrib import github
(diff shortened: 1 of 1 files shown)


ppanero:
@kpsherva It sounds good to me. I was not happy with the original names, this are taken from ILS (as suggested by @ntarocco). I do like the icons suggestions, wonder how understandable would it be for "dates" (the rest I think is clear), since is not lexicographical nor numeric order.

kpsherva (approved):

PR #223, opened 2023-03-15
config cleanup

ppanero:
I'm sorry, I could not help myself. In the comments I will leave extra things that should be refactored.

It is failing in isort/pydocstyle, I will fixed if the changes are agreed upon.

### DIFF
--- a/invenio.cfg
+++ b/invenio.cfg
@@ -20,34 +20,11 @@ from invenio_records_resources.services.records.queryparser import (
     QueryParser,
     SearchFieldTransformer,
 )
+from invenio_oauthclient.views.client import auto_redirect_login
 
-# Imprint CF
-from invenio_rdm_records.contrib.imprint.custom_fields import (
-    IMPRINT_NAMESPACE,
-    IMPRINT_CUSTOM_FIELDS,
-    IMPRINT_CUSTOM_FIELDS_UI
-)
+from zenodo_rdm.custom_fields import CUSTOM_FIELDS_UI, CUSTOM_FIELDS, NAMESPACES
 
-# Journal CF
-from invenio_rdm_records.contrib.journal.custom_fields import (
-    JOURNAL_NAMESPACE,
-    JOURNAL_CUSTOM_FIELDS,
-    JOURNAL_CUSTOM_FIELDS_UI
-)
 
-# MEETING CF
-from invenio_rdm_records.contrib.meeting.custom_fields import (
-    MEETING_NAMESPACE,
-    MEETING_CUSTOM_FIELDS,
-    MEETING_CUSTOM_FIELDS_UI
-)
-
-# Thesis CF
-from invenio_rdm_records.contrib.thesis.custom_fields import (
-    THESIS_NAMESPACE,
-    THESIS_CUSTOM_FIELDS,
-    THESIS_CUSTOM_FIELDS_UI
-)
 
 # Flask
 # =====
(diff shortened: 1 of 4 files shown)


site/zenodo_rdm/config.py
ppanero, line 1: Modules have a config **entrypoint** now. Most of the `ZENODO_` configuration (i.e. not overwriting other RDM modules) should be moved here.

site/zenodo_rdm/config.py
```diff
@@ -163,228 +176,20 @@
 #: Email title template.
 SUPPORT_EMAIL_CONFIRM_TITLE_TEMPLATE = "zenodo_rdm/email_confirm_title.html"
 
+# Search query of recent uploads
+# Defaults to newest records search
+ZENODO_FRONTPAGE_RECENT_UPLOADS_QUERY = "/api/records?sort=newest&size=10"
```
ppanero, line 181: There is no `ext.py` in the `zenodo_rdm` module. Therefore, there is no need to prefix this variables. If we go for prefixing for distintcion, the `SUPPORT` ones should be `ZENODO_SUPPORT_...`

invenio.cfg
ppanero, line 1: IMO the `ZENODO_LEGACY_SEARCH_MAP` and `RDM_SEARCH` logic is big enough to be moved to a `search.py` in the `zenodo_rdm` module. However, I did not touch it because there is work going on there and I did not want to cause conflicts.

site/zenodo_rdm/redirector.py
```diff
@@ -0,0 +1,221 @@
...
+# Copyright (C) 2023 CERN.
+#
+# ZenodoRDM is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Redirector functions and rules."""
+
+from flask import request, url_for
+
+ZENODO_TYPE_SUBTYPE_LEGACY = {
```
ppanero, line 12: This could be a config and then use `current_app.get()` but since it was already _not_ being used... I didnt

site/zenodo_rdm/custom_fields.py
```diff
@@ -0,0 +1,154 @@
...
+
+from invenio_records_resources.services.custom_fields import (
+    BaseCF,
+    DoubleCF,
+    ISODateStringCF,
+    KeywordCF,
+    TextCF,
+)
+from invenio_rdm_records.contrib.imprint import (
+    IMPRINT_NAMESPACE,
```
ppanero, line 18: This is for rdm-records, but wouldn't it make more sense to remove prefixes of the actual custom field? (i.e. `NAMESPACE, CUSTOM_FIELDS`, etc.). Why:

- Shortens the names
- It does not add value to the name, since we are importing from the actual package, i.e. already in the `.contrib.imprint...`
- If there are collisions they can always be imported with `as ...`

site/zenodo_rdm/redirector.py
```diff
@@ -0,0 +1,221 @@
...
+
+    The following routes are redirected as follows:
+        - /communities/about/<id>/ -> GET /communities/<pid_value>
+        - /collection/user-<id> -> GET /communities/<pid_value>
+        - /communities/<community_id>/about -> GET /communities/<pid_value>
+
+    :return: url for the view 'invenio_app_rdm_communities.communities_detail'
+    :rtype: str
+    """
+    _id = request.view_args.get("id", request.view_args.get("community_id"))
```
ntarocco, line 57: Don't you want to make the `community_id` mandatory? That `.get` can be `None`?
```suggestion
    _id = request.view_args.get("id", request.view_args["community_id"])
```
Comment valid for all the methods below.

ntarocco (approved):

site/zenodo_rdm/redirector.py
ppanero: I had only moved code, not evaluated its correctness. Thanks! I've modified the access :)

Issue #201, opened 2023-02-27
migration: custom fields


kpsherva:
Prerequisites:

https://github.com/inveniosoftware/invenio-app-rdm/issues/2019

- [ ] add missing custom fields to zenodo-rdm (from invenio-rdm-records/invenio-app-rdm)
- [ ] make sure the fields are migrated

PR #224, opened 2023-03-17
Migrator custom fields

ppanero:
- closes https://github.com/zenodo/zenodo-rdm/issues/201
- closes https://github.com/zenodo/zenodo-rdm/issues/207

# IMPORTANT

Tests will fail until the release of the above module since `custom_fields` won't tranformed.

# Question

- Do all the `dwc:`, `codemeta:`, etc. related fields need migration? IIUC they are new fields to help with query use cases like _Zenodeo_, right?
    - Some like `rightsHolder` could potentially be infered
    - Many of the dwc could be migrated from the `notes` field. However, the processing would be very error prone. Is this something we want to pursue?


site/zenodo_rdm/migrator/transform/records.py
```diff
@@ -0,0 +1,206 @@
...
+                "session_part": metadata.get("meeting", {}).get("session_part"),
+                "session": metadata.get("meeting", {}).get("session"),
+                "title": metadata.get("meeting", {}).get("title"),
+                "url": metadata.get("meeting", {}).get("url"),
+            },
+            "imprint:imprint": {
+                "isbn": metadata.get("imprint", {}).get("isbn"),
+                "publisher": metadata.get("imprint", {}).get("publisher"),
+                "place": metadata.get("imprint", {}).get("place"),
+                # QUESTION: the following two are not found in Zenodo
```
ppanero, line 134: @alejandromumo I'm a bit delayed in catching up with this. Are these fields new, or the match any old Zenodo ones?

site/zenodo_rdm/migrator/transform/records.py
```diff
@@ -0,0 +1,206 @@
...
+                # "title": metadata.get("journal", {}).get("title"),
+                # "pages": metadata.get("journal", {}).get("volume"),
+            },
+            "thesis:university": metadata.get("thesis", {}).get("university"),
+            # QUESTION: migrate Zenodo thesis.supervisors as record collaborators?
```
ppanero, line 139: @slint @zzacharo What do we do about Zenodo's `thesis.supervisors`. They do not map to any new field, however we could move them to collaborators (not sure which role)

site/zenodo_rdm/migrator/transform/records.py
alejandromumo: They match Zenodo's :

```
"part_of.title": "custom_fields.imprint:imprint.title",
"part_of.pages": "custom_fields.imprint:imprint.pages",
```

See this record on [Sandbox](https://sandbox.zenodo.org/api/records/1174530). 
It has all the `book` fields and these two (`title` and `pages`) are set to `part_of`

site/zenodo_rdm/migrator/transform/records.py
alejandromumo: We removed the supervisors from the thesis field for now, given that `supervisors` can be retrieved from the contributors whose role is `Supervisor`. However, that supervisor would only be the thesis supervisor if the resource type is `Publication/Thesis`. But correct me if I'm, wrong  @slint

ppanero: @alejandromumo I agree with that, **new** records in ZenodoRDM will have supervisors in the contributors field. However, my comment was on the reverse order about **migrating** the old Zenodo ones (that are already in `thesis.supervisors`), which the zenodo_rdm.migrator currently simply discards (i.e. we are loosing metadata).

site/zenodo_rdm/migrator/transform/records.py
ppanero: Fixed, Thanks a lot!

.github/workflows/tests.yml
```diff
@@ -63,6 +63,7 @@ jobs:
       - name: Install dependencies
         run: |
           pipenv install --dev --deploy --system
+          pip install "./site[$EXTRAS]"
```
slint, line 66:
```suggestion
          pip install "./site[tests]"
```

minor: since we're in the tests CI, I think that's reasonable to hardcode (and ideally we don't want to add any extras in the future)

site/tests/migrator/transform/test_record_transform.py
slint, line 1: Nice! Thanks a lot for adding tests for the migration :)

note: we should also look into doing the same for the `rdm-migrator` module.

site/zenodo_rdm/migrator/transform/records.py
slint: So, the rule is that we'll just migrate these into the `contributors` field (which is not yet implemented in this class, but has to).

Something to keep in mind is that we should keep the legacy REST API and the migrated metadata in sync, i.e. if someone submits `thesis.supervisors` via the legacy REST API, it should end up in `contributors`

.github/workflows/tests.yml
ppanero: While I do agree, the `$EXTRAS` variable is already defined and it includes the search extra (opensearch). At the moment the are not being used (I think), but the `run-tests.sh` script boots up the containers anyway. options:
a) simplify tests run, and remove containers etc. However, this would make it inconsistent with other modules. It would reduce running time (not much though)
b) keep extras as is

If a), I would suggest to be tackeld in a different issue.

site/zenodo_rdm/migrator/transform/records.py
ppanero: Perfect, I've added a light version of contributors migration. For the moment only takes care of adding the supervisors (which is what concerns this PR) but sets the logistics to allow adding contributors already.

site/tests/migrator/transform/test_record_transform.py
ppanero: https://github.com/inveniosoftware/invenio-rdm-migrator/issues/38

site/zenodo_rdm/migrator/transform/records.py
```diff
@@ -0,0 +1,229 @@
...
+            "files": "public" if is_open else "restricted",
+        }
+        if record["access_right"] == "embargoed":
+            r["embargo"] = {
+                "until": record["embargo_date"],
+                "active": True,
+            }
+        return r
+
+    def _metadata(self, entry):
```
ppanero, line 72: note: this looks like could be extrated to something like `MetadataEntry` class. However, out of scope for this PR.

site/tests/migrator/transform/test_record_transform.py
```diff
@@ -0,0 +1,280 @@
...
+import pytest
+
+from zenodo_rdm.migrator.transform.records import (
+    ZenodoRecordEntry,
+    ZenodoRecordTransform,
+)
+
+
+@pytest.fixture(scope="module")
+def zenodo_record_data():
```
kpsherva, line 20: 👍

site/zenodo_rdm/migrator/transform/records.py
```diff
@@ -0,0 +1,229 @@
...
+                "place": metadata.get("imprint", {}).get("place"),
+                "title": metadata.get("part_of", {}).get("title"),
+                "pages": metadata.get("part_of", {}).get("pages"),
+            },
+            "thesis:university": metadata.get("thesis", {}).get("university"),
+        }
+
+        return self._drop_nones(cf)
+
+    def _drop_nones(self, d):
```
kpsherva, line 167: nit: could be a utility for migration, probably will be needed in the future in other places

slint (approved):
Nice! We should try to keep these to a minimum in general, but I think once we progress through the milestones it'll become easier.

site/zenodo_rdm/migrator/transform/records.py
zzacharo: it would be nice to address also this https://github.com/inveniosoftware/invenio-rdm-migrator/issues/36

ppanero: @zzacharo if it is within the scope I can do it while tackling now the inclusion requests. Should not take too much time I think, no?

zzacharo: it shouldn't but it has some impact of course in cds-rdm...but I think it will clean up the interface and improve readability :)

Issue #210, opened 2023-03-03
OAI-PMH: Fix email


lnielsen:
https://zenodo-rdm.web.cern.ch/oai2d?verb=Identify

![Screenshot 2023-03-03 at 17 45 12](https://user-images.githubusercontent.com/1698163/222778010-def18a4a-0a0d-4bda-ac95-ec1d512f680f.png)

info@inveniosoftware.org -> info@zenodo.org

PR #227, opened 2023-03-20
config: Update OAI-PMH adminEmail

jennur:
Closes https://github.com/zenodo/zenodo-rdm/issues/210

This PR adds the correct `adminEmail` to `OAISERVER_ADMIN_EMAILS` for Zenodo.

Screenshot 2023-03-20 at 10 42 48

### DIFF
--- a/invenio.cfg
+++ b/invenio.cfg
@@ -133,6 +133,7 @@ DATACITE_ENABLED = True
 DATACITE_PREFIX = "10.5281"
 DATACITE_TEST_MODE = True
 OAISERVER_ID_PREFIX = "zenodo-rdm.web.cern.ch"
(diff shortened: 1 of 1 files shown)


ppanero (approved):

kpsherva:
actually, a bit hasty here. Isn't it helm chart value rather than exposed email on GH?

slint:
We have the same [on the legacy codebass](https://github.com/zenodo/zenodo/blob/d03140f91987e1bf22bcb3240f244ab985d9ecd3/zenodo/config.py#L98), since this is public information anyways (i.e. it's going to be displayed on the OAI-PMH response, and it's also the same address that we use for support).

PR #232, opened 2023-03-20
Migrator contributors

ppanero:
- includes #224 **review only the last commit**
- closes https://github.com/zenodo/zenodo-rdm/issues/231

### DIFF
--- a/site/zenodo_rdm/migrator/transform/records/metadata.py
+++ b/site/zenodo_rdm/migrator/transform/records/metadata.py
@@ -1,0 +1,98 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 CERN.
+#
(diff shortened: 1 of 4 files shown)


site/zenodo_rdm/migrator/transform/records/metadata.py
```diff
@@ -0,0 +1,96 @@
...
+"""Zenodo migrator metadata entry transformer."""
+
+from invenio_rdm_migrator.transform import Entry
+from nameparser import HumanName
+
+
+class ZenodoMetadataEntry(Entry):
+    """Metadata entry transform."""
+
+    @classmethod
```
ppanero, line 17: all methods could be classmethod or even staticmethod, should we change the interface? I do not see a point where there is class/instance state/attributes to be accessed since this are very strict hardcoded migration rules.

PR #234, opened 2023-03-21
migrator: add simple custom fields

### LINKED ISSUE #233: migrate dwc custom fields
codemeta is new, therefore there is nothing to migrate

ppanero:
- closes #233


site/tests/migrator/transform/test_record_transform.py
```diff
@@ -136,6 +136,28 @@ def zenodo_record_data():
                     }
                 ],
             },
+            "custom": {
+                # dwc
+                "dwc:collectionCode": ["MUSM", "ZMPC"],
```
ppanero, line 141: Note: I have added tests for the most common fields that I found among several taxonomic treatments in Zenodo. Is it worth adding for every single field since the logic is the same?

kpsherva: if the logic is the same, probably not. However if you want the migration to be airtight, I would add it for the peace of mind :)

kpsherva (approved):

slint (approved):

Issue #213, opened 2023-03-03
Link redirection: Export links does not redirect


lnielsen:
Export links like below does not work:
https://www.zenodo.org/record/7696674/export/hx

https://github.com/zenodo/rdm-project/issues/17

jennur:
Missing 

BibTex,
DCAT,
JSON-LD (missing serializer in `invenio-rdm-records`)

PR #235, opened 2023-03-21
config: add redirect rules for legacy export endpoints

jennur:
Closes https://github.com/zenodo/zenodo-rdm/issues/213

This PR adds redirect rules for legacy export endpoints for Zenodo, including

- hx -> bibtex
- dcite4 -> datacite-xml
- xd -> dublincore
- geojson -> GeoJSON
- xm -> marcxml
- dcat -> DCAT-AP


site/zenodo_rdm/redirector.py
```diff
@@ -165,6 +174,25 @@ def record_export_view():
...
+
+    The following routes are redirected as follows:
+        -  /records/<pid_value>/export/<legacy_export_format> -> GET /record/<pid_value>/export/<export_format>
+
+    :return: url for the view 'invenio_app_rdm_records.record_export'
+    :rtype: str
+    """
+
+    args = request.view_args
+    export_format = request.path.split("/")[-1]
```
ppanero, line 188: is the format also available in the view arguments? i.e. `args["export_format"]`

jennur: No, as I had to add the full path in the config like this `/records/<pid_value>/export/hx`, otherwise I got into a redirect loop!

site/zenodo_rdm/redirector.py
```diff
@@ -165,6 +174,25 @@ def record_export_view():
...
+    legacy_value = ZENODO_RECORD_EXPORTERS_LEGACY.get(export_format)
```
alejandromumo, line 189: out of curiosity, what would happen with invalid formats? Does the new endpoint handle it the same way as the previous?

site/zenodo_rdm/redirector.py
alejandromumo: Shouldn't the format be available in `request.args` or `request.view_args`?  Like [here](https://github.com/zenodo/zenodo-rdm/blob/08d21c983ec5218bc42a7e990db01369798029b9/site/zenodo_rdm/redirector.py#L102)

site/zenodo_rdm/redirector.py
jennur: If the format is invalid, it just goes to 404. But in this case, I have specified this function only for the valid ones (for the reason explained above).

site/zenodo_rdm/config.py
```diff
@@ -272,6 +273,30 @@
...
+        "target": legacy_record_export_view,
+    },
+    "redirect_legacy_record_export_view_xm": {
+        "source": "/records/<pid_value>/export/xm",
+        "target": legacy_record_export_view,
+    },
+    "redirect_legacy_record_export_view_dcat": {
+        "source": "/records/<pid_value>/export/dcat",
+        "target": legacy_record_export_view,
+    },
```
slint, line 299: minor: it feels a bit duplicated to have both the `ZENODO_RECORD_EXPORTERS_LEGACY` config, which contains the mapping, and these entries. Would it make sense to "generate" the dictionary entries from config:

```python
**{
    f"redirect_leagcy_record_export_view_{key}": {
        "source": f"/records/<pid_value>/export/{key}"
        "target": legacy_record_export_view,
    } for key in ZENODO_RECORD_EXPORTERS_LEGACY
}
```

...or make some sort of utility function generator, to avoid having `ZENODO_RECORD_EXPORTERS_LEGACY`:

```python
**{
    f"redirect_leagcy_record_export_view_{key}": {
        "source": f"/records/<pid_value>/export/{key}"
        "target": legacy_record_export_view(key, val),  # this would return a function that does the mapping
    } for key, val in.items()
}
```

jennur:
## Updates

Implemented the function as suggested by @slint : 
```python
 **{
    f"redirect_legacy_record_export_view_{key}": {
        "source": f"/records/<pid_value>/export/{key}"
        "target": legacy_record_export_view,
    } for key in ZENODO_RECORD_EXPORTERS_LEGACY
}
```

slint (approved):

PR #253, opened 2023-04-03
badges: ui implementation for badges in zenodo

TLGINO:
" added suport for:
  - F1000
  - Dryad
  - Github
  - OpenAire
  - Orcid
  - Reana
  - Inspire
 * closes https://github.com/zenodo/zenodo-rdm/issues/249


site/zenodo_rdm/utils.py
```diff
@@ -0,0 +1,134 @@
...
+# under the terms of the MIT License; see LICENSE file for more details.
+"""Utility functions."""
+
+
+def github_link_render(record):
+    """Entry for Github."""
+    # [TODO] Integrate Inveio_Github here
+    resources = []
+
+    related_identifiers = record["ui"].get("related_identifiers", [])
```
anikachurilova, line 15: Little (maybe stupid) comment/quiestion: do we know for sure that `record["ui"]` exists?
Is it better to use `lodash/get` here instead?

TLGINO: from tests and debugging it looks to me like `ui` always exists although this doesn't seem to be the case for `related_identifiers` (hence the `get`).
But you're right using `get` for both is probably safer

site/zenodo_rdm/utils.py
```diff
@@ -0,0 +1,134 @@
...
+    return resources or None
+
+
+def openaire_link_render(record):
+    """Entry for Openaire."""
+    resources = []
+
+    related_identifiers = record.get("ui", {}).get("related_identifiers", [])
+    for identifier in related_identifiers:
+        if identifier["scheme"] == "url" and "openaire" in identifier["identifier"]:
```
kpsherva, line 35: missing top level domain for openaire? ping @slint

slint:
Screenshot:

image

We're missing an icon for the Journal of Brief Ideas, but can be done in a separate issue.

Issue #241, opened 2023-03-29
Data migration: Record version relationships


kpsherva:
`RDMVErsionState` is already being migrated however we are missing the drafts

ppanero:
**PIDs**

There are 4 tables related to pids:
- pidstore_pid
- pidstore_recid
- pidstore_redirect
- pidrelations_pidrelation

The first one (`pidstore_pid`) is being migrated as part of the `records` stream. The
stream is creating the records `recid`, `doi` and `oai`, and the parents' `recid`
(equivalent to the concept recid). Missing:
- Parent `doi` (i.e. concept DOI)

`pidstore_recid` is a simple counter for the recid value, since that has changed in RDM
it does not need to be migrated. `pidrelations_pidrelation` is removed from RDM,
relationships are handled via de `rdm_versions_state` table. This last one is migrated
by the `records` stream. Missing:
- `next_draft`, deposits (i.e. drafts)

The complete `pidstore_redirect` is missing migration. Migration should be a 1-to-1?

**Records**

There are several tables related to records:

- records_metadata
- records_metadata_version
- transactions
- All the tables related to invenio-sipstore

`records_metadata` the extract query in ZenodoRDM only includes active children pids.
Parent records are migrated from the `conceptrecid` inside the metadata. Missing:
- Deposits/Drafts
- Deleted records

`records_metadata_version`, a decission is needed here. Do we want to migrate as is, and
then only transform if we need to recover? or potentially even migrate the versions
asynchronously (not during the main migration downtime).

`transactions` is not being migrated, is needed for auditing and produced by sql
continuum. Do we want to migrate it, and keep it? (not used atm in RDM).

`records_buckets` will be left as part of the files migration.

A decission is needed to integrate invenio-sipstore and the migrate the content of
its tables. If needed.

ppanero:
drafts are good to go, they have 2 things missing:
- files (actual files, and bucket creation if they are new)
- we need a record cache to:
     - reuse the uuid of the record if the draft is of an existing record
     - to avoid yielding a PersistentIdentifier row if it already exists

PR #256, opened 2023-04-11
drafts and versions

ppanero:
- closes https://github.com/zenodo/zenodo-rdm/issues/241

### DIFF
--- a/site/zenodo_rdm/migrator/extract.py
+++ b/site/zenodo_rdm/migrator/extract.py
@@ -100,6 +100,23 @@ EXTRACT_RECORDS_SQL = """
     ) TO STDOUT;
 """
 
+EXTRACT_DRAFTS_SQL = """
+COPY (
+    SELECT row_to_json(deposits)
+    FROM (
+        SELECT
+            r.*
+        FROM records_metadata as r
+            JOIN pidstore_pid
+                ON pidstore_pid.object_uuid = r.id
+        WHERE
+            pidstore_pid.pid_type = 'depid' AND
+            pidstore_pid.status = 'R' AND
+            pidstore_pid.object_type = 'rec'
+    ) as deposits
+) TO STDOUT;
+"""
+
 EXTRACT_REQUESTS_SQL = """
 COPY (
     SELECT row_to_json(requests)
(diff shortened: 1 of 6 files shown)


Pipfile.lock
alejandromumo, line 1: This was updated on master and has conflicts. 
is this the version I've sent you?

site/zenodo_rdm/migrator/streams.yaml
```diff
@@ -1,36 +1,36 @@
...
-  load:
-    db_uri: postgresql://zenodo:zenodo@localhost:5432/zenodo
-    tmp_dir: /path/to/tmp
-communities:
-  extract:
-    filepath: /path/to/communities.jsonl
-  load:
-    db_uri: postgresql://zenodo:zenodo@localhost:5432/zenodo
-    tmp_dir: /path/to/tmp
+# users:
```
zzacharo, line 1: leftover, you need to revert it back to its initial state

site/zenodo_rdm/migrator/extract.py
```diff
@@ -100,6 +100,25 @@ def run(self):
     ) TO STDOUT;
 """
 
+EXTRACT_DRAFTS_SQL = """
```
ppanero, line 103: this will be updated by alex

site/zenodo_rdm/migrator/transform/records.py
```diff
@@ -10,17 +10,17 @@
...
 
 class ZenodoRecordTransform(RDMRecordTransform):
     """Zenodo to RDM Record class for data transformation."""
 
     def _communities(self, entry):
-        communities = entry["json"].get("communities")
-        if communities:
-            slugs = [slug for slug in communities]
-            return {"ids": slugs, "default": slugs[0]}
+        # communities = entry["json"].get("communities")
```
zzacharo, line 20: can we put this back or there is something wrong with drafts that belong already to communities?

site/zenodo_rdm/migrator/extract.py
slint: No, the one here is correct, it's the same I have :)

PR #257, opened 2023-04-17
migrator: added multiple communities.

alejandromumo:
closes https://github.com/zenodo/zenodo-rdm/issues/229

### DIFF
--- a/site/tests/migrator/transform/test_community_transform.py
+++ b/site/tests/migrator/transform/test_community_transform.py
@@ -1,0 +1,91 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 CERN.
+#
+# ZenodoRDM is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Test community transform for RDM migration."""
+import dictdiffer
+import pytest
+
+from zenodo_rdm.migrator.transform.communities import (
+    ZenodoCommunityEntry,
+    ZenodoCommunityMemberEntry,
+)
+
+
+@pytest.fixture(scope="module")
+def zenodo_community_data():
+    """Full Zenodo community as a dictionary."""
+    return {
+        "created": "2023-01-01 12:00:00.00000",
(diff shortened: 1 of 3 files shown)


site/zenodo_rdm/migrator/transform/records/records.py
```diff
@@ -83,13 +83,12 @@ def _custom_fields(self, entry):
...
-    def _community_id(self, entry):
+    def _communities(self, entry):
         communities = entry["json"].get("communities")
         if communities:
+            slugs = [slug for slug in communities]
             # TODO: handle all slugs
-            slug = communities[0]
-            if slug:
-                return {"ids": [slug], "default": slug}
+            return {"ids": slugs, "default": slugs[0]}  # TODO default?
```
alejandromumo, line 91: To be discussed: should the default community be the first one from the list of the record's communities? The notion of `default` did not exist on zenodo so we need to decide what do we want as default. 

I can't remember of better suggestions except doing something like "prioritizing". E.g: prioritize some communities like Biodiversity Literature Repository, LORY, etc. If not found, default to the first one from the list.

ppanero (approved):

site/tests/migrator/transform/test_community_transform.py
```diff
@@ -0,0 +1,91 @@
...
+@pytest.fixture(scope="module")
+def expected_rdm_community():
+    """Expected community in zenodo-rdm."""
+    return {
+        "created": "2023-01-01 12:00:00.00000",
+        "updated": "2023-01-31 12:00:00.00000",
+        "version_id": 1,
+        "slug": "migrator",
+        "json": {
+            "files": {"enabled": True},
```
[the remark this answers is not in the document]
kpsherva: I will add it to my list of questions for the current scope :)

site/zenodo_rdm/migrator/transform/records/records.py
alejandromumo: Good tip, thanks!

https://github.com/zenodo/zenodo-rdm/issues/259

PR #263, opened 2023-04-21
WIP: added missing fields.

### LINKED ISSUE #262: Migration: record non-required fields
Currently records migration do not consider non-required fields. 

In this issue we should migrate them. if possible. In case any field can't be migrated now, we should open an issue and tackle it separately. 

--- 

**Note on vocabularies**

Some fields are values from controlled vocabularies (e.g. licenses). Since both zenodo-rdm and inveniordm have vocabularies, its values should be resolved and validated.  Moreover, in Zenodo we have "aliases" for some of these vocabularies (e.g. licenses). 

Therefore, we must migrate the vocabularies themselves before migrating records and provide some kind of compatibility for the aliases.

This issue is out of scope from this task, for now, and should be tackled separately. That said, in this issue we can assume that the vocabularies are already migrated (e.g. no resolving nor validation is needed for now)

alejandromumo:
closes #262

### DIFF
--- a/site/zenodo_rdm/migrator/transform/communities.py
+++ b/site/zenodo_rdm/migrator/transform/communities.py
@@ -28,6 +28,6 @@ class ZenodoCommunityTransform(CommunityTransform):
 
     def _featured_community(self, entry):
         """Transform the featured community."""
-        if entry["is_featured"]:
+        if entry.get("is_featured", False):
             return ZenodoFeaturedCommunityEntry().transform(entry)
         return {}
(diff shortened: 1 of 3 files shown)


site/zenodo_rdm/migrator/transform/communities.py
```diff
@@ -113,6 +113,6 @@ def _community_members(self, entry):
 
     def _featured_community(self, entry):
         """Transform the featured community."""
-        if entry["is_featured"]:
+        if entry.get("is_featured", False):
```
zzacharo, line 116: according to the extraction query isn't the "is_featured" property always there?

zzacharo (approved):
Needs rebasing and squashing the commits but it looks good!

zzacharo:
I am merging this and tests should be fixed in the next PR when finishing the missing fields of locations and dates!

PR #273, opened 2023-04-26
Query fixes and dump docs

slint:
- migrator: add and fix extract SQL queries
- migrator: document how to generate dumps

### DIFF
--- a/site/zenodo_rdm/migrator/README.md
+++ b/site/zenodo_rdm/migrator/README.md
@@ -129,3 +129,38 @@ current_rdm_records_service.indexer.bulk_index((rec.id for rec in records))
   user doesn't seem to have the correct roles when visiting a community, then you might
   need to run the `site.zenodo_rdm.migrator.utils.invalidate_user_community_roles_cache`
   so users community roles are being re-cached.
+
+## Generating dumps
+
+In order to generate the dumps used by the migrator, you need access to a legacy Zenodo
+database with either live data or a snapshot.
+
+In `zenodo-rdm/site/zenodo_rdm/migrator/extract.py`, you'll find the SQL queries for
+dumping the data needed for each of the migration streams. You can run these queries
+directly on the legacy database, in the following manner:
+
+```shell
+# NOTE: Adjust the connection string as necessary
+DB_URI="postgresql://zenodo@zenodo-legacy-db-host:5432/zenodo"
+
+# Users, ~30min
+psql $DB_URI -f users_dump.sql | sed 's/\\\\/\\/g' | gzip -c > "users-$(date -I).jsonl.gz"
+# Communities, ~5min
+psql $DB_URI -f communities_dump.sql | sed 's/\\\\/\\/g' | gzip -c > "communities-$(date -I).jsonl.gz"
+# Community record inclusion requests, ~10min
+psql $DB_URI -f requests_dump.sql | sed 's/\\\\/\\/g' | gzip -c > "requests-$(date -I).jsonl.gz"
+# Records, ~2-3h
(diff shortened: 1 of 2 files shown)


site/zenodo_rdm/migrator/README.md
```diff
@@ -129,3 +129,38 @@ current_rdm_records_service.indexer.bulk_index((rec.id for rec in records))
...
   need to run the `site.zenodo_rdm.migrator.utils.invalidate_user_community_roles_cache`
   so users community roles are being re-cached.
+
+## Generating dumps
+
+In order to generate the dumps used by the migrator, you need access to a legacy Zenodo
+database with either live data or a snapshot.
+
+In `zenodo-rdm/site/zenodo_rdm/migrator/extract.py`, you'll find the SQL queries for
+dumping the data needed each of the migration streams. You can run them directly on the
```
ppanero, line 139:
```suggestion
dumping the data needed for each of the migration streams. You can run them directly on the
```

site/zenodo_rdm/migrator/README.md
```diff
@@ -129,3 +129,38 @@ current_rdm_records_service.indexer.bulk_index((rec.id for rec in records))
...
+
+# Users, ~30min
+psql $DB_URI -f users_dump.sql | sed 's/\\\\/\\/g' | gzip -c > "users-$(date -I).jsonl.gz"
+# Communities, ~5min
+psql $DB_URI -f communities_dump.sql | sed 's/\\\\/\\/g' | gzip -c > "communities-$(date -I).jsonl.gz"
+# Community record inclusion requests, ~10min
+psql $DB_URI -f requests_dump.sql | sed 's/\\\\/\\/g' | gzip -c > "requests-$(date -I).jsonl.gz"
+# Records, ~2-3h
+psql $DB_URI -f records_dump.sql | sed 's/\\\\/\\/g' | gzip -c > "records-$(date -I).jsonl.gz"
+# Deposits/drafts, ~30min
```
ppanero, line 154: Is this still the case after the query was fixed?

site/zenodo_rdm/migrator/extract.py
```diff
@@ -23,8 +23,6 @@ def run(self):
         """Yield one element at a time."""
         with open(self.filepath, "r") as reader:
             for line in reader:
-                # TODO JSONl file should be cleaned up before loading
-                line = line.strip().replace("\\\\", "\\")
                 yield json.loads(line)
```
ppanero, line 26: Does it still need to be loaded or can we use `jsonlines.read` now?

zzacharo: Are we sure we don't need still to use `.strip()`?

slint: > Does it still need to be loaded or can we use `jsonlines.read` now?

I think that's an [additional dependency](https://jsonlines.readthedocs.io/en/latest/), not sure if/where it's defined, and looks like there aren't many of its features we'd be using.

> Are we sure we don't need still to use `.strip()`?

Looks like it's not needed for `json`:

```python
In [1]: import json
In [2]: json.loads(" {} ")
Out[2]: {}
```

site/zenodo_rdm/migrator/README.md
slint: Yes, surprisingly... Now that I think of it, maybe it's because I had run a similar query beforehand, so there was some caching? On the other hand, compared to the Records query, it has one less `JOIN` (since we don't use `pidrelations`), and that might have an impact, especially since there are no indices on some of the joined columns.

zzacharo:
I am merging this and we fix tests in an upcoming PR

Issue #271, opened 2023-04-25
Permissions: introduce a permission for restricted files


anikachurilova:
Moving code from `invenio-rdm-records` [PR](https://github.com/inveniosoftware/invenio-rdm-records/pull/1280) as this is only Zenodo-related stuff. Left PR open for now for saving comments history reason. Will be closed afterwards.

Override `can_read_files` permission policy (see: https://github.com/inveniosoftware/invenio-rdm-records/pull/1280#discussion_r1176451624)
 
- `can_community_access = False`:
    - record owner - access
    - community owner - no access
    - random user - no access
- `can_community_access = True`:
    - record owner - access
    - community owner - access
    - random user - no access

See: https://github.com/inveniosoftware/invenio-rdm-records/issues/1274

PR #274, opened 2023-04-27
permissions: override RDM permission policy

anikachurilova:
* manage permissions depending on a new permission flag
* closes https://github.com/zenodo/zenodo-rdm/issues/271

### DIFF
--- a/site/tests/conftest.py
+++ b/site/tests/conftest.py
@@ -12,13 +12,26 @@ from collections import namedtuple
 import pytest
 from flask_security import login_user
 from flask_security.utils import hash_password
+from invenio_access.models import ActionRoles
+from invenio_access.permissions import superuser_access, system_identity
+from invenio_accounts.models import Role
 from invenio_accounts.testutils import login_user_via_session
+from invenio_administration.permissions import administration_access_action
 from invenio_app.factory import create_api as _create_api
+from invenio_communities import current_communities
+from invenio_communities.communities.records.api import Community
+from invenio_pidstore.errors import PIDDoesNotExistError
+from invenio_vocabularies.proxies import current_service as vocabulary_service
+from invenio_vocabularies.records.api import Vocabulary
+
+from zenodo_rdm.permissions import ZenodoRDMRecordPermissionPolicy
 
 
 @pytest.fixture(scope="module")
 def app_config(app_config):
     """Mimic an instance's configuration."""
+    app_config["RDM_PERMISSION_POLICY"] = ZenodoRDMRecordPermissionPolicy
+    app_config["REST_CSRF_ENABLED"] = False
     return app_config
 
 
@@ -32,8 +45,10 @@ RunningApp = namedtuple(
     "RunningApp",
     [
         "app",
+        "superuser_identity",
         "location",
         "cache",
+        "resource_type_v",
     ],
 )
 
@@ -41,8 +56,10 @@ RunningApp = namedtuple(
 @pytest.fixture
 def running_app(
     app,
+    superuser_identity,
     location,
     cache,
+    resource_type_v,
 ):
     """Fixture provides an app with the typically needed db data loaded.
 
@@ -51,8 +68,10 @@ def running_app(
     """
     return RunningApp(
         app,
+        superuser_identity,
         location,
         cache,
+        resource_type_v,
     )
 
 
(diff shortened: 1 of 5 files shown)


site/tests/conftest.py
```diff
@@ -84,3 +109,303 @@ def client_with_login(client, users):
...
+    # Properties
+    #
+    @property
+    def user(self):
+        """Get the user."""
+        return self._user
+
+    @property
+    def email(self):
+        """Get the user."""
```
TLGINO, line 168: very minor: wrong getter description for `email` and `password`

site/zenodo_rdm/legacy/generators.py
```diff
@@ -0,0 +1,29 @@
...
+    """Conditional generator for files restriction for community."""
+
+    def __init__(self, then_, else_):
+        """Constructor."""
+        super().__init__("files", then_, else_)
+
+    def _condition(self, record, **kwargs):
+        """Check if community can access restricted files of the migrated record."""
+        # can_community_read_files = getattr(record.parent, permission_flags.can_community_read_files, True)
+        can_community_read_files = False
```
jrcastro2, line 23: Leftover?
```suggestion
        can_community_read_files = getattr(record.parent, permission_flags.can_community_read_files, True)
```

site/zenodo_rdm/config.py
```diff
@@ -312,3 +313,6 @@
 }
 
 REDIRECTOR_RULES.update(EXPORT_REDIRECTS)
+
+# Permissions
+RDM_PERMISSION_POLICY = ZenodoRDMRecordPermissionPolicy
```
jrcastro2, line 318: Question: Why do we need to set here the `RDM_PERMISSION_POLICY` when it was already set in the `invenio.cfg`?

site/tests/conftest.py
```diff
@@ -84,3 +109,303 @@ def client_with_login(client, users):
...
+        """Logout the client."""
+        res = client.get(f"{base_path}logout")
+        assert res.status_code < 400
+        return client
+
+
+@pytest.fixture(scope="session")
+def UserFixture():
+    """Class to create user fixtures from."""
+    return UserFixture_
```
jrcastro2, line 224: This was already moved to pytest-invenio, see [here](https://github.com/inveniosoftware/pytest-invenio/blob/master/pytest_invenio/fixtures.py#L855). I think it can be removed.

site/tests/test_record_file_permissions.py
```diff
@@ -0,0 +1,207 @@
...
+    response = client.post(
+        f"/records/{recid}/draft/files/test.pdf/commit", headers=headers
+    )
+    assert response.status_code == 200
+
+    # Publish it
+    response = client.post(f"/records/{recid}/draft/actions/publish", headers=headers)
+    assert response.status_code == 202
+
+    return recid
```
jrcastro2, line 46: This looks like a copy of [this](https://github.com/inveniosoftware/invenio-rdm-records/blob/36a0abfb68a4a4e0ce2208087840de448a06c523/tests/resources/test_record_file_permissions.py#L23) from invenio-rdm-records, can we reuse it instead?

site/tests/conftest.py
```diff
@@ -8,17 +8,36 @@
...
 
 @pytest.fixture(scope="module")
 def app_config(app_config):
     """Mimic an instance's configuration."""
+    supported_configurations = ["RDM_PERMISSION_POLICY"]
+
+    for config_key in supported_configurations:
+        app_config[config_key] = getattr(config, config_key, None)
+
+    app_config["REST_CSRF_ENABLED"] = False
```
jrcastro2, line 40: leftover?

site/zenodo_rdm/legacy/generators.py
anikachurilova: Not the leftover, will be uncommented and tested after that field is added: https://github.com/inveniosoftware/invenio-rdm-records/pull/1290

site/tests/test_record_file_permissions.py
anikachurilova: I'm not sure i understand how.. unless we move it to `pytest-invenio`?

jrcastro2: Right, currently it doesn't seem that we can reuse utility functions from tests. We would probably need to group these utility functions per repo and make them available for others to be used. It looks like a bigger task, out of the scope of this task I would say.

Moving this to `pytest-invenio` doesn't seem possible because this is specific to invenio RDM.

site/zenodo_rdm/legacy/generators.py
```diff
@@ -0,0 +1,34 @@
...
+        try:
+            can_community_read_files = dict_lookup(
+                record.parent, "permission_flags.can_community_read_files"
+            )
+        except KeyError:
+            can_community_read_files = True
+
+        is_restricted = super()._condition(record, **kwargs)
+        if is_restricted:
+            return not can_community_read_files
```
ppanero, line 32: I'm not quite sure I understand why the `not` is needed. Thinking of the following case:

- The record has the `can_community_read_files` flag set to `False`. Then `can_community_read_files=False`.
- The files are restricted, therefore the code will enter the `if` in line 31. And will return `not can_community_read_files`, which is `not False`, therefore returning `True`. 

I might have misunderstood something on the flow, but I think is wrong and could lead to data leak. Shouldn't the return's be like:

```python
return can_community_read_files if restricted else False
```

Meaning: if files are restricted return what the flag has as value, otherwise default to false

site/tests/conftest.py
```diff
@@ -84,3 +103,214 @@ def client_with_login(client, users):
     login_user(user)
     login_user_via_session(client, email=user.email)
     return client
+
+
+@pytest.fixture(scope="module")
+def resource_type_v(app):
```
ppanero, line 109: most of these are copy pasted from rdm-records/app-rdm records. It makes me wonder if we should have sort of an RDM specific subset of pytets fixtures. cc @zzacharo @kpsherva, it might be worth discussing this at some point.

ppanero (commented):
some tests are failing because it gives a 200 when it should be a 403, not a 100% sure but I think is due to the issue I pointed out above.

anikachurilova:
> some tests are failing because it gives a 200 when it should be a 403, not a 100% sure but I think is due to the issue I pointed out above.

@ppanero it is failing because [this](https://github.com/inveniosoftware/invenio-rdm-records/pull/1290) is not merged yet and it puts the default of `True` to the flag

site/zenodo_rdm/legacy/generators.py
anikachurilova: In case of `can_community_read_files = False`, the condition of class `IfFilesRestrictedForCommunity` will return true, which means that files are restricted for community (as we can understand from the name of the class).

So in case of the returning value being `True`, the permission `can_read_files_owner` will be used. So in the end if the files are restrcited and `can_community_read_files = False` will lead to the permission logic where community can not see files.

Probably the naming is confusing, but for the better understanding you can check the tests (hope it helps).

site/zenodo_rdm/legacy/permissions.py
slint, line 1: minor/nit: this file (and `generators.py`) can go one level higher in the folder structure, i.e. `site/zenodo_rdm/permissions.py`, since they don't have anything specific to legacy. We also try to keep the `legacy` folder kind of isolated, so that it only contains legacy REST API logic (to make it easy to remove once we deprecate things).

slint (approved):
LGTM, just a minor/nit move. 

I think the dependent PR is merged, but there's a release train going on, so I guess we just wait for it :)

PR #275, opened 2023-04-27
migration: remove draft communities

ppanero:
- closes https://github.com/zenodo/zenodo-rdm/issues/241

Cases tested:

- New draft: can view, save. cannot publish due to missing files
- Published record: can view
- A record with 2 published versions and a draft of the first one. Can view all, if edit on v1 leads to draft v1
    - v1 - draft v1
    - v2
- A published record with a draft for a new version. Can view all, if new version on v1 leads to draft v2
    - v1
    - draft v2

### DIFF
--- a/invenio.cfg
+++ b/invenio.cfg
@@ -232,7 +232,7 @@ RDM_SEARCH = {
     ],
     "query_parser_cls": QueryParser.factory(
         mapping=ZENODO_LEGACY_SEARCH_MAP,
-        tree_transformer_factory=SearchFieldTransformer,
+        tree_transformer_cls=SearchFieldTransformer,
     ),
 }
 


site/zenodo_rdm/migrator/transform/records.py
```diff
@@ -76,9 +76,13 @@ def _transform(self, entry):
         is_draft = entry["json"]["_deposit"]["status"] == "draft"
 
         if is_draft:
+            # FIXME: draft communities could be review or addition
+            # we might need to differentiate those
+            parent = (self._parent(entry),)
+            parent["communities"] = {}
```
alejandromumo, line 82: Is this intended? Just double checking

alejandromumo (commented):
LGTM, left a minor comment

site/zenodo_rdm/migrator/transform/records.py
ppanero: Nice catch, half of it yes, half no 😂

PR #276, opened 2023-04-27
migrator: added dates record.

alejandromumo:
Tested record with `dates` and they were succesfully migrated and render in the UI.

### DIFF
--- a/site/zenodo_rdm/migrator/transform/entries/records/metadata.py
+++ b/site/zenodo_rdm/migrator/transform/entries/records/metadata.py
@@ -277,57 +277,23 @@ class ZenodoRecordMetadataEntry(Entry):
             "creators": cls._creators(entry["creators"]),
             "contributors": contributors,
             "publisher": entry.get("imprint", {}).get("publisher"),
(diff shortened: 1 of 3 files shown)


site/zenodo_rdm/migrator/transform/entries/records/metadata.py
```diff
@@ -313,9 +313,7 @@ def transform(cls, entry):
 
         dates = cls._dates(entry.get("dates"))
         if dates:
-            # TODO uncomment after tested
-            # metadata.update({"dates": dates})
-            pass
+            metadata.update({"dates": dates})
```
ppanero, line 316: i added `return drop_nones(metadata)` at the end of the drafts entry. I wonder if it would be better to do the same for records. Would fix the custom fields empty keys issues and also save `if` clauses. WDYT?

alejandromumo: Yeah that's exactly the issue I was trying to bypass. So I agree with your suggestion

PR #277, opened 2023-04-27
migration: remove demo data (users, communities, records)

ppanero:


wipe_recreate.sh
```diff
@@ -38,9 +38,10 @@ invenio communities custom-fields init
 
 # Add demo and fixtures data
 # -------------
-invenio rdm-records fixtures
-invenio rdm-records demo
+invenio rdm-records fixture
+psql << "DELETE FROM accounts_user WHERE id=1;"  # remove admin@inveniosoftware.org
```
ppanero, line 42: How do we automate this?

slint: Let's address tomorrow :)

PR #283, opened 2023-04-28
migrator: added metadata locations for records.

### LINKED ISSUE #262: Migration: record non-required fields

alejandromumo:
closes #262 

This PR fixes minor issues and adds tests for `metadata` fields.

### DIFF
--- a/migrate.sh
+++ b/migrate.sh
@@ -1,3 +1,11 @@
+##############
+#
+# Initial notes, read first.
+#
+# -  To be able to see deposits (with admin user), the owner of records must be hardcoded to the admin user id.
+#
+##############
+
 # Clear DB
 # Only run if DB has data.
 invenio shell --no-term-title -c "import redis; redis.StrictRedis.from_url(app.config['CACHE_REDIS_URL']).flushall(); print('Cache cleared')"
@@ -24,8 +32,9 @@ invenio communities custom-fields init
 python -m zenodo_rdm.migrator site/zenodo_rdm/migrator/streams.yaml
 
 # Update ID sequences in DB
-psql -h <host> -U zenodo -d zenodo
+# This is developed programatically, not need to run it manually.
 <<com
+psql -h <host> -U zenodo -d zenodo
 SELECT MAX(id) from accounts_user;
 ALTER SEQUENCE ALTER SEQUENCE accounts_user_id_seq RESTART WITH {ID}; # Insert ID from before
 com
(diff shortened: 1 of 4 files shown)


migrate.sh
```diff
@@ -24,6 +32,7 @@ invenio communities custom-fields init
 python -m zenodo_rdm.migrator site/zenodo_rdm/migrator/streams.yaml
 
 # Update ID sequences in DB
+# TODO this is being developed programatically. Might not be needed at all.
```
ppanero, line 35: I think we can remove this since there is the alter sequence PR. Or at least fully comment it.

migrate.sh
ppanero, line 1: cant comment on the line:
- the fixutres command creates a user that we need to remove, with id 1... How do we do that? we discussed it briefly (it was not solved by adding an empty users.yaml, since it is created programmatically)

site/zenodo_rdm/migrator/transform/entries/records/metadata.py
```diff
@@ -287,8 +285,7 @@ def transform(cls, entry):
             ),
             "references": cls._references(entry.get("references")),
             "dates": cls._dates(entry.get("dates")),
-            # TODO: fix location entry
-            # "locations": cls._locations(entry.get("locations")),
+            "locations": cls._locations(entry.get("locations")),
```
ppanero, line 288: Needs to be added to drafts too (and tests also adapted). Either in this PR or in https://github.com/zenodo/zenodo-rdm/pull/286 

I think we can merge this and tackle changes in mine? WDYT

ppanero (commented):
move to approved after answering comments! nice jobe 🚀

site/zenodo_rdm/migrator/stream.py
```diff
@@ -74,6 +73,8 @@
 
 # TODO: move this to a proper place or support custom stream classes
```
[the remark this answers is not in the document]
alejandromumo: Created an issue for this to be reviewed by @zzacharo  https://github.com/zenodo/zenodo-rdm/issues/290

migrate.sh
alejandromumo: Not sure what the consensus was on this one. Ping @zzacharo @slint

site/zenodo_rdm/migrator/transform/entries/records/metadata.py
alejandromumo: Added!

PR #285, opened 2023-05-02
migrator: fix users dump query

slint:


site/zenodo_rdm/migrator/extract.py
```diff
@@ -53,7 +53,7 @@ def run(self):
             coalesce(user_i.identities, null) AS identities,
             coalesce(user_sa.session_activity, null) AS session_activity
         FROM accounts_user AS u
-        JOIN userprofiles_userprofile up ON u.id = up.user_id
+        LEFT JOIN userprofiles_userprofile up ON u.id = up.user_id
```
slint, line 56: If only I had paid attention in school...

site/zenodo_rdm/migrator/transform/entries/users.py
```diff
@@ -57,9 +57,11 @@ def _displayname(self, entry):
 
     def _profile(self, entry):
         """Returns the profile."""
-        return {
-            "full_name": entry.get("full_name"),
-        }
+        res = {}
```
slint, line 60: I don't think this is necessary, but if we don't have a "full name", the profile should be an empty dict by default.

Issue #228, opened 2023-03-20
legacy API: support contributor, journal, imprint, thesis, conference, and notes fields


slint:
- [x] `contributors`
- [x] Journal
  - [x] `journal_title`
  - [x] `journal_volume`
  - [x] `journal_issue`
  - [x] `journal_pages`
- [x] Conference
  - [x] `conference_title`
  - [x] `conference_acronym`
  - [x] `conference_dates`
  - [x] `conference_place`
  - [x] `conference_url`
  - [x] `conference_session`
  - [x] `conference_session_part`
- [x] Imprint
  - [x] `imprint_publisher`
  - [x] `imprint_isbn`
  - [x] `imprint_place`
  - [x] `partof_title`
  - [x] `partof_pages`
- [x] Thesis
  - [x] `thesis_supervisors`
    - This should be serialized into `contributors` with `contributorType: Supervisor`
  - [x] `thesis_university`

Goal is that Dryad, Plazi and ZHB can submit their records using the existing API with all supported metadata fields.

PR #289, opened 2023-05-03
Add legacy metadata fields

alejandromumo:
closes https://github.com/zenodo/zenodo-rdm/issues/228


site/zenodo_rdm/legacy/deserializers/schemas.py
ppanero, line 1: Why deserializers are not Marshmallow but just dict operations? should we make it consistent?

site/zenodo_rdm/legacy/deserializers/schemas.py
```diff
@@ -9,6 +9,7 @@
 
 from datetime import date
 
+from invenio_records.dictutils import clear_none
```
ppanero, line 12: should open an issue, we have plenty of `drop_nones` usage. Very similar if not identical in functionality:
- https://github.com/inveniosoftware/invenio-rdm-migrator/blob/bd88485a7cb3d09a69d2b802e0029082dc232188/invenio_rdm_migrator/transform/base.py#LL36C1-L36C1
- https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/dictutils.py#L14

ppanero (approved):

ppanero:
@slint The relevant issue is still "in progress" but I think this PR is completed. Merge?

site/zenodo_rdm/legacy/deserializers/schemas.py
slint: When I started this, I had a lot of trouble making a Marshmallow schema for loading that made sense. Most fields ended up being implemented via `fields.Method`, which defeated the point of trying to formalize the process...

We could revisit in the future, the RO-Crate deserializer was done via marshmallow: https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/resources/deserializers/rocrate/schema.py, and though it suffers from the same issue with using too often `fields.Method`, it plays better with error formatting, etc.

site/zenodo_rdm/legacy/serializers/schemas.py
```diff
@@ -63,6 +171,20 @@ def dump_resource_type(self, result, original, **kwargs):
...
         return result
 
+    @post_dump(pass_original=True)
+    def dump_contributor(self, result, original, **kwargs):
+        """Dumps contributor."""
+        contributors = result.get("contributors", [])
+        new_contributors = []
+        for c in contributors:
+            contributor_type = c.get("type")
+            # Supervisors are serialized in "thesis", thus are removed from contributors list
```
slint, line 181: A bit of a bamboozle, but in legacy you can also have normal contributors that are supervisors :)

- [`contributors.type:Supervisor`](https://zenodo.org/search?page=1&size=20&q=contributors.type:Supervisor)
- [`contributors.type:Supervisor AND _exists_:thesis`](https://zenodo.org/search?page=1&size=20&q=contributors.type:Supervisor%20AND%20_exists_:thesis): this is even worse, basically there's duplication between the fields, and sometimes even mismatches...

I can't think of an elegant solution for handling this and preserve the user input... Maybe the solution that is "consistent" with the reality is to serialize everything as contributors (and never return `thesis.supervisors` basically).

slint (approved):

site/zenodo_rdm/legacy/deserializers/schemas.py
zzacharo: I saw that one in invenio-rdm-migrator but I didn't change it because I assumed that we want ideally not to depend on `invenio-records`....

ppanero: Yes, dependency is kind of an issue, I dont know what we should do... in the end this module is `invenio-` so it should be ok to depend. On the other hand it would make tests slower if we bring in app fixtures and co. Im conflicted. To be evaluated.

site/zenodo_rdm/legacy/serializers/schemas.py
```diff
@@ -63,6 +155,16 @@ def dump_resource_type(self, result, original, **kwargs):
                 result[f"{upload_type}_type"] = resource_type_id.split("-")[-1]
         return result
 
+    @post_dump(pass_original=True)
+    def dump_notes(self, result, original, **kwargs):
+        """Dump notes."""
+        additional_descriptions = original.get("additional_descriptions", [])
+        breakpoint()
```
slint, line 162:
```suggestion
```

Time for a Kit-Kat 🍫

slint (approved):

PR #291, opened 2023-05-04
Files stream, constraints/indices recreation, and minor fixes

slint:
- migrator: files stream fixes and integration
- migrator: drafts transformer fixes
- migrator: update migrate steps
- migrator: constraint and index loading scripts
- migrator: handle missing conceptrecid for drafts


zzacharo (approved):

ppanero (approved):
approving to avoid blocking. Would it be worth to have them done programmatically (even if its an `execute("SQL STATEMENT")`) to avoid human errors due to missing steps?

PR #292, opened 2023-05-04
legacy api: de/serializer for funding field

### LINKED ISSUE #265: legacy API: grant reference ID aliases
In Zenodo records metadata, we have references to the same licenses with different IDs (e.g. `cc0`, `cc-zero`, `cc0-1.0`, etc.). These are "aliases" for the same Creative Commons Zero v1.0 license, and are implemented via PIDs that have their object_uuid point to the same "record". The way forward is to:

- Provide similar compatibility in RDM with these "aliases". This is also important so that we have legacy REST API compatibility (i.e. a user submitting a record with a license using one of the alias IDs)
- That means that we have to also import the license vocabulary (or at least the ID aliases) before importing the records, so that the references are valid when we load the

---

The same issue applies to grants/awards references, but in that case, we have completely different identifiers, since in RDM we migrated from FundRef IDs to using ROR for funders, which affects the grant/award identifier (`10.13039/100000001::1540608` vs. `021nxhr62::1540608`)... This needs a bit more thought in general so that we can minimize also how many data entries we need to keep in the DB/search.

ppanero:
- closes #265 

Created a desposit using legacy API, see request and response:

Screenshot 2023-05-05 at 17 24 10

The same deposit is displayable in the web interface (i.e. RDM API)

Screenshot 2023-05-05 at 17 11 38

### DIFF
--- a/site/zenodo_rdm/legacy/deserializers/schemas.py
+++ b/site/zenodo_rdm/legacy/deserializers/schemas.py
@@ -11,6 +11,39 @@ from datetime import date
 
 from nameparser import HumanName
 
+# based on
+# DOI to id https://digital-repositories.web.cern.ch/zenodo/support/operations/#openaire-grants-import
+# id to ROR: https://github.com/inveniosoftware/invenio-vocabularies/blob/master/invenio_vocabularies/config.py#L64
+FUNDER_DOI_TO_ROR = {
+    "10.13039/501100001665": "00rbzpz17",
+    "10.13039/501100002341": "05k73zm37",
+    "10.13039/501100000923": "05mmh0f86",
+    "10.13039/100018231": "01gavpb45",
+    "10.13039/501100000024": "00k4n6c32",
+    "10.13039/501100000780": "00k4n6c32",
+    "10.13039/501100000806": "02k4b9v70",
+    "10.13039/501100001871": "00snfqn58",
+    "10.13039/501100002428": "013tf3c58",
+    "10.13039/501100004488": "03n51vw80",
+    "10.13039/501100004564": "01znas443",
+    "10.13039/501100000925": "011kf5r70",
+    "10.13039/100000002": "01cwqze88",
+    "10.13039/501100000038": "01h531d29",
+    "10.13039/100000001": "021nxhr62",
+    "10.13039/501100003246": "04jsz6e67",
+    "10.13039/501100000690": "10.13039/501100000690",
+    "10.13039/100014013": "001aqnf71",
+    "10.13039/501100001602": "0271asj38",
+    "10.13039/501100001711": "00yjd3n13",
+    "10.13039/100001345": "006cvnv84",
+    "10.13039/501100004410": "04w9kkr77",
+    "10.13039/100004440": "029chgv08",
+    "10.13039/501100006364": "03m8vkq32",
+    "10.13039/501100000780": "00k4n6c32",
+}
+
+FUNDER_ROR_TO_DOI = {v: k for k, v in FUNDER_DOI_TO_ROR.items()}
+
 
 class LegacyRecordTransform:
     """Transform a Zenodo legacy record to an RDM record."""
@@ -38,6 +71,28 @@ class LegacyRecordTransform:
             ret.append(r)
         return ret
 
+    def _funding(self, grants):
+        """Transform funding information."""
+        _funding = []
+        for grant in grants:
+            # format "10.13039/501100000780::190101904"
+            id_parts = grant["id"].split("::", 1)
+            assert len(id_parts) == 2
+
+            # special case of funder loaded in Zenodo with no DOI related to it.
+            # it is part of the vocabulary.
+            funder_doi = id_parts[0]
+            award_id = id_parts[1]
+            funder_doi_or_ror = FUNDER_DOI_TO_ROR.get(funder_doi, funder_doi)
+
+            _funding.append(
+                {
+                    "funder": {"id": funder_doi_or_ror},
+                    "award": {"id": f"{funder_doi_or_ror}::{award_id}"},
+                }
+            )
+        return _funding
+
     def _metadata(self, data):
         """Transform the metadata of a record."""
         if data:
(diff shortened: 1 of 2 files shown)


site/zenodo_rdm/legacy/deserializers/schemas.py
```diff
@@ -38,6 +73,72 @@ def _creators(self, data):
...
+        if id_:  # should be part of vocabulary
+            # format `funder_id::code` as per
+            # invenio_vocabularies/contrib/awards/datastreams.py#L53
+            return {"id": id_}
+
+        # custom award
+        return {
+            "title": grant["title"],
+            "number": grant["code"],
+            "identifiers": [
```
ppanero, line 121: We might have to configure [VOCABULARIES_AWARD_SCHEMES](https://github.com/inveniosoftware/invenio-vocabularies/blob/master/invenio_vocabularies/config.py#L58). At the momtne it contains only DOI and URL, however legacy records would contain OAF, pURL and eurepo.

site/zenodo_rdm/legacy/deserializers/schemas.py
```diff
@@ -11,6 +11,41 @@
...
+    "10.13039/100014013": "001aqnf71",
+    "10.13039/501100001602": "0271asj38",
+    "10.13039/501100001711": "00yjd3n13",
+    "10.13039/100001345": "006cvnv84",
+    "10.13039/501100004410": "04w9kkr77",
+    "10.13039/100004440": "029chgv08",
+    "10.13039/501100006364": "03m8vkq32",
+}
+
+# FIXME: missing on one or the other side of the data source
```
ppanero, line 44: What do we do with these three? At the moment they would just be let pass as actual vocabularies with the corresponding DOI (not ror). But not as custom funder. The last 2 cases (taraexp and chistera) would be pass as custom funder, since they have no DOI.

site/zenodo_rdm/legacy/serializers/schemas.py
```diff
@@ -63,6 +66,90 @@ def dump_resource_type(self, result, original, **kwargs):
                 result[f"{upload_type}_type"] = resource_type_id.split("-")[-1]
         return result
 
+    def _funder(self, funder):
+        """Serialize RDM funder into Zenodo legacy funder."""
+        legacy_funder = {"name": funder["name"]}
+
+        f_identifiers = {}
+        for scheme, identifier in funder.get("identifiers").items():
+            if scheme == "oaf":
```
ppanero, line 75: We might need to modify [VOCABULARIES_FUNDER_SCHEME](https://github.com/inveniosoftware/invenio-vocabularies/blob/master/invenio_vocabularies/config.py#L49). At the moment it does not accept oaf.

site/zenodo_rdm/legacy/serializers/schemas.py
```diff
@@ -63,6 +66,92 @@ def dump_resource_type(self, result, original, **kwargs):
...
+            legacy_grant["title"] = title
+        except StopIteration:
+            pass
+
+        value = award.get("acronym")
+        if value:
+            legacy_grant["acronym"] = value
+
+        a_identifiers = {}
+        for identifier in award.get("identifiers"):
```
ppanero, line 108: We might have to configure [VOCABULARIES_AWARD_SCHEMES](https://github.com/inveniosoftware/invenio-vocabularies/blob/master/invenio_vocabularies/config.py#L58). At the momtne it contains only DOI and URL, however legacy records would contain OAF, pURL and eurepo.

site/zenodo_rdm/legacy/serializers/schemas.py
```diff
@@ -63,6 +66,96 @@ def dump_resource_type(self, result, original, **kwargs):
...
+        if value:
+            legacy_funder["country"] = value
+
+        return legacy_funder
+
+    def _award(self, award):
+        """Serialize an RDM award into a legacy Zenodo grant."""
+        legacy_grant = {
+            "code": award["number"],
+            "internal_id": f"{award['funder']['id']}::{award['id']}",
```
ppanero, line 95: This is part of the questions of the issue:
a) do we change format (break backwards compatibility) and do ror::award-id
b) we can implement an inverse dict and translate to funder doi. If this is the case, what do we do with those new ones not in the mapping?

site/zenodo_rdm/legacy/deserializers/schemas.py
```diff
@@ -38,6 +73,26 @@ def _creators(self, data):
             ret.append(r)
         return ret
 
+    def _funding(self, grants):
+        """Transform funding information."""
+        _funding = []
+        for grant in grants:
+            # format "10.13039/501100000780::190101904"
+            id_parts = grant["id"].split("::")
```
slint, line 81:
```suggestion
            id_parts = grant["id"].split("::", 1)
```

minor: just in case there's a grant that has for some reason a code with `::` (i.e. something like `10.1234/foobar::grant::code`)

site/zenodo_rdm/legacy/deserializers/schemas.py
```diff
@@ -38,6 +73,26 @@ def _creators(self, data):
...
+            assert len(id_parts) == 2
+
+            # special case of funder loaded in Zenodo with no DOI related to it.
+            # it is part of the vocabulary.
+            funder_id = FUNDER_DOI_TO_ROR.get(id_parts[0], id_parts[0])
+
+            _funding.append(
+                {
+                    "funder": {"id": funder_id},
+                    "award": {"id": id_parts[1]},
```
slint, line 91:
```suggestion
                    "award": {"id": id_parts[0 ] + "::" + id_parts[1]},
```

minor: I think `award.id`s are the fully formatted funder + code string (see example in [reference docs](https://inveniordm.docs.cern.ch/reference/metadata/#funding-references-0-n)).

site/zenodo_rdm/legacy/serializers/schemas.py
slint: `b)` is the way, we reverse map, and in case there's no mapping we maintain the RDM-style i.e. `ror-funder::award-code`

site/zenodo_rdm/legacy/serializers/schemas.py
```diff
@@ -63,6 +66,96 @@ def dump_resource_type(self, result, original, **kwargs):
...
+        }
+
+        # if english does not exist default to the first key
+        # FIXME: should default to default locale?
```
slint, line 99: English is fine, it's also what we had in legacy. Actually to be more accurate we don't know if it's always English, but we also don't know the language of the imported strings :)

slint (commented):
Couple of comments/clarifications and minor fixes, but otherwise looks good :+1:

slint (approved):

PR #293, opened 2023-05-04
files: add direct file stream loading from csv files

zzacharo:
requires https://github.com/inveniosoftware/invenio-rdm-migrator/pull/60


migrate.sh
```diff
@@ -25,40 +22,22 @@ invenio index queue init purge
...
-
-
-# Restore FK/PK/unique constraints and indices
-psql $DB_URI -f create_constraints.sql
-psql $DB_URI -f restore_indices.sql
+python -m zenodo_rdm.migrator site/zenodo_rdm/migrator/tmp/10k/streams.yaml
 
 # Update ID sequences in DB
-psql $DB_URI -f update_sequences.sql
+# This is developed programatically, not need to run it manually.
```
ppanero, line 35: The post_load should have updated all sequences. For documentation purposes is okay, but it looks very much like a command to run. If we want to explain the steps maybe we should start an actual documentation page for it?

migrate.sh
```diff
@@ -25,40 +22,22 @@ invenio index queue init purge
...
-invenio files location create --default 'default-location' /opt/invenio/var/instance/data
+invenio files location create --default 'default-location'  $(pipenv run invenio shell --no-term-title -c "print(app.instance_path)")'/data'
+# invenio files location create --default 'default-location' /opt/invenio/var/instance/data
 invenio index init --force
 invenio rdm-records custom-fields init
 invenio communities custom-fields init
 
-# Backup FK/PK/unique constraints
-cd scripts
-DB_URI="postgresql://zenodo:zenodo@localhost/zenodo"
```
ppanero, line 36: why is all the contraints work form alex being removed?

site/zenodo_rdm/migrator/load.py
```diff
@@ -0,0 +1,93 @@
...
+import psycopg
+from invenio_rdm_migrator.load.postgresql import _ts
+from invenio_rdm_migrator.streams.files.models import (
+    FilesBucket,
+    FilesInstance,
+    FilesObjectVersion,
+)
+
+
+class ZenodoFilesLoad:
```
ppanero, line 22: Can extend invenio-rdm-migrator `Load` class (similar constructor). Then overrite `_load` and `run` to skip the prepare step.

site/zenodo_rdm/migrator/load.py
```diff
@@ -0,0 +1,93 @@
...
+                conn.commit()
+
+    def run(self, entries, cleanup=False):
+        """Load entries."""
+        from datetime import datetime
+
+        start_time = datetime.now()
+        print(f"Stream started {start_time.isoformat()}")
+
+        self._load()
```
ppanero, line 88: Should call `post_load` afterwards (i.e. super()) and reuse form parent base Load class.

site/zenodo_rdm/migrator/stream_definition.py
```diff
@@ -0,0 +1,42 @@
...
+
+from invenio_rdm_migrator.extract import Extract
+from invenio_rdm_migrator.streams import StreamDefinition
+from invenio_rdm_migrator.transform import Transform
+
+
+class ZenodoDirectLoadStreamDefition(StreamDefinition):
+    """Zenodo to RDM direct loading class i.e no ET steps."""
+
+    class NoExtract(Extract):
```
ppanero, line 17: You can put those two calsses as root (not inner calsses). Theren there would be no need to overwrite and could use `StreamDefinition` directly (i.e. simplifying the __main__)

site/zenodo_rdm/migrator/load.py
```diff
@@ -0,0 +1,93 @@
...
+        Loads the tables in the order given by the generator.
+        """
+
+        def get_cols(model):
+            from dataclasses import fields
+
+            return ", ".join([f.name for f in fields(model)])
+
+        with psycopg.connect(self.db_uri) as conn:
+            for model in self.models:
```
slint, line 43: shelve: from a quick look this is the only part that seems to be different from the original `PostgreSQLCopyLoad`. Could we abstract this somehow to avoid the duplication? Or move the inner loop into a separate underscore-method?

site/zenodo_rdm/migrator/stream_definition.py
```diff
@@ -0,0 +1,42 @@
...
+
+        def _transform(self, entry):
+            """Transform entry."""
+            yield entry
+
+    def __init__(self, name, load_cls):
+        """Constructor."""
+        super(ZenodoDirectLoadStreamDefition, self).__init__(
+            name=name,
+            extract_cls=self.NoExtract,
```
slint, line 36:
```suggestion
            extract_cls=self.NoExtract,
```

shelve: alternative would me to not pass anything (and default to this), but this requires changes to the base class. Definitely something that we could move in `rdm-migrator`.

slint (approved):
LGTM, two things to shelve and revisit later.

site/zenodo_rdm/migrator/stream_definition.py
zzacharo: I didn't want that just because the DirectLoad makes sense only if you pass these dummy internal classes...

migrate.sh
slint: I found out kind of late that [`_post_load` is never called anywhere](https://github.com/inveniosoftware/invenio-rdm-migrator/blob/bd88485a7cb3d09a69d2b802e0029082dc232188/invenio_rdm_migrator/load/postgresql.py#L134). I then used the query to find the sequences and produced the `update_sequences.sql`.

site/zenodo_rdm/migrator/load.py
```diff
@@ -0,0 +1,58 @@
...
+        """Constructor."""
+        super().__init__(
+            db_uri=db_uri,
+            table_generators=[ZenodoFilesTableGenerator()],
+            tmp_dir=tmp_dir,
+        )
+        # override default as we don't generate the table files again
+        self.tmp_dir = Path(tmp_dir)
+
+    def _validate(self):
```
ppanero, line 51: mental note: all load implementations either return True either pass... we can remove the `abstractmethod` decorator in base?

ppanero (approved):

PR #295, opened 2023-05-04
migrate: fix path resolotion to related files

zzacharo:


migrate.sh
```diff
@@ -47,20 +47,23 @@ psql $DB_URI -f drop_indices.sql
 
 
 # Run migration
+cd -
```
slint, line 50: Doesn't this go to the "previous" directory? I think if you want to assume the parent folder of this script file you have to do `script_path=$(dirname "$0")` and use that (maybe it's a good idea in general to build our paths around this assumption/point?)

migrate.sh
```diff
@@ -24,45 +24,50 @@ invenio index queue init purge
 # NOTE: db init is not needed since DB keeps being created
 #       Just need to create all tables from it.
 invenio db create
-# TODO: add back when pipenv access problem is fixed
-#invenio files location create --default 'default-location'  $(pipenv run invenio shell --no-term-title -c "print(app.instance_path)")'/data'
+# TODO: add back when pipenv access problem is fixed. Use this locally!!!
+# invenio files location create --default 'default-location'  $(pipenv run invenio shell --no-term-title -c "print(app.instance_path)")'/data'
```
slint, line 28:
```suggestion
invenio files location create --default "default-location" $(invenio shell --no-term-title -c 'print(f"{app.instance_path}/data")'
```

This should also work, and then we can also uncomment the line completely and remove the one below. Checked on an OpenShift pod, and the output is the same:

```bash
[1011760000@web-debug src]$ invenio shell --no-term-title -c 'print(f"{app.instance_path}/data")'
/opt/invenio/var/instance/data
```

slint (approved):
LGTM! Probably needs a rebase though after the wiping protection PR :)

migrate.sh
zzacharo: Removed it!

Issue #284, opened 2023-04-28
Create a new request type


anikachurilova:
```
LegacyRecordUpgrade

type = "legacy-record"
topic = record
created_by = community (?) system (?)
receiver = record owner
```

See notes: https://github.com/inveniosoftware/invenio-rdm-records/issues/1274
RFC: https://codimd.web.cern.ch/kvN6rZAhTrOCzzeOdzX7cQ#
EPIC: https://github.com/zenodo/zenodo-rdm/issues/248

PR #318, opened 2023-05-08
requests: add a legacy record upgrade request

anikachurilova:
* add submit, accept, decline and expire actions logic
* closes https://github.com/zenodo/zenodo-rdm/issues/284

The description of a request preview:
Screenshot 2023-05-11 at 17 49 56

### DIFF
--- a/site/tests/conftest.py
+++ b/site/tests/conftest.py
@@ -29,6 +29,7 @@ from invenio_vocabularies.proxies import current_service as vocabulary_service
 from invenio_vocabularies.records.api import Vocabulary
 
 from zenodo_rdm.custom_fields import CUSTOM_FIELDS, CUSTOM_FIELDS_UI, NAMESPACES
+from zenodo_rdm.legacy.requests.record_upgrade import LegacyRecordUpgrade
 from zenodo_rdm.permissions import ZenodoRDMRecordPermissionPolicy
 
 
@@ -43,6 +44,7 @@ def app_config(app_config):
     app_config["RDM_CUSTOM_FIELDS"] = CUSTOM_FIELDS
     app_config["RDM_CUSTOM_FIELDS_UI"] = CUSTOM_FIELDS_UI  #  UI components
     app_config["RDM_PERMISSION_POLICY"] = ZenodoRDMRecordPermissionPolicy
+    app_config["REQUESTS_REGISTERED_TYPES"] = [LegacyRecordUpgrade()]
     return app_config
 
 
@@ -497,10 +499,10 @@ def minimal_record():
         "pids": {},
         "access": {
             "record": "public",
-            "files": "public",
+            "files": "restricted",
         },
         "files": {
-            "enabled": False,  # Most tests don't care about files
+            "enabled": True,  # Most tests don't care about files
         },
         "metadata": {
             "creators": [
@@ -533,6 +535,18 @@ def minimal_community():
             "title": "Biodiversity Literature Repository",
             "type": {"id": "topic"},
         },
+    }
+
+
+@pytest.fixture()
+def minimal_community2():
+    """Data for a minimal community two."""
+    return {
+        "slug": "rdm",
+        "access": {
+            "visibility": "public",
+        },
+        "metadata": {"title": "Research Data Management", "type": {"id": "topic"}},
     }
 
 
@@ -670,8 +684,32 @@ def community(running_app, community_type_record, community_owner, minimal_commu
 
 
 @pytest.fixture()
+def community2(running_app, community_type_record, community_owner, minimal_community2):
+    """Get the current RDM records service."""
+    return _community_get_or_create(minimal_community2, community_owner.identity)
+
+
+@pytest.fixture()
 def community_with_uploader_owner(
     running_app, community_type_record, uploader, minimal_community
 ):
     """Create a community with an uploader owner."""
     return _community_get_or_create(minimal_community, uploader.identity)
+
+
+@pytest.fixture(scope="function")
+def superuser(app, db, UserFixture, superuser_role_need):
+    """Superuser fixture."""
+    u = UserFixture(
+        email="superuser@inveniosoftware.org",
+        password="superuser",
+    )
+    u.create(app, db)
+
+    datastore = app.extensions["security"].datastore
+    _, role = datastore._prepare_role_modify_args(u.user, "superuser-access")
+
+    datastore.add_role_to_user(u.user, role)
+    db.session.commit()
+
+    return u
--- a/site/tests/requests/test_record_upgrade_request.py
+++ b/site/tests/requests/test_record_upgrade_request.py
@@ -39,7 +39,6 @@ def _send_post_action(action, request_id, client, headers, expected_status_code)
 def create_record_w_file(client, record, headers):
     """Create record with a file."""
     # Create draft
-    record["files"] = {"enabled": True}
     response = client.post("/records", json=record, headers=headers)
     assert response.status_code == 201
     recid = response.json["id"]
--- a/site/tests/requests/test_record_upgrade_request.py
+++ b/site/tests/requests/test_record_upgrade_request.py
@@ -1,0 +1,957 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 CERN.
+#
+# Zenodo is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Test Zenodo legacy record upgrade request."""
+from io import BytesIO
+
+import pytest
+from invenio_access.permissions import system_identity
+from invenio_rdm_records.proxies import current_rdm_records
+from invenio_rdm_records.records import RDMParent, RDMRecord
+from invenio_records_resources.services.errors import PermissionDeniedError
+from invenio_requests import current_requests_service
+
+from zenodo_rdm.legacy.requests.utils import submit_record_upgrade_request
+
+
+@pytest.fixture()
+def service(running_app):
+    """The record service."""
+    return current_rdm_records.records_service
+
+
+def _send_post_action(action, request_id, client, headers, expected_status_code):
+    """Send http-post request to perform an action on a request"""
+    response = client.post(
+        f"/requests/{request_id}/actions/{action}",
+        headers=headers,
+        json={},
+    )
+
+    assert response.status_code == expected_status_code
+    return response
+
+
+def create_record_w_file(client, record, headers):
+    """Create record with a file."""
+    # Create draft
+    record["files"] = {"enabled": True}
+    response = client.post("/records", json=record, headers=headers)
+    assert response.status_code == 201
+    recid = response.json["id"]
+
+    # Attach a file to it
+    response = client.post(
+        f"/records/{recid}/draft/files", headers=headers, json=[{"key": "test.pdf"}]
+    )
+    assert response.status_code == 201
+    response = client.put(
+        f"/records/{recid}/draft/files/test.pdf/content",
+        headers={
+            "content-type": "application/octet-stream",
+            "accept": "application/json",
+        },
+        data=BytesIO(b"testfile"),
+    )
+    assert response.status_code == 200
+    response = client.post(
+        f"/records/{recid}/draft/files/test.pdf/commit", headers=headers
+    )
+    assert response.status_code == 200
+
+    # Publish it
+    response = client.post(f"/records/{recid}/draft/actions/publish", headers=headers)
+    assert response.status_code == 202
+
+    return recid
+
+
+def _add_permission_flags(permission_flags, record, db):
+    db_record = RDMRecord.get_record(record._record.id)
+    db_record.parent.permission_flags = permission_flags
+    db_record.parent.commit()
+    db.session.commit()
+
+
+def _create_legacy_record(identity, record, permission_flags, db, service):
+    """Create and publish a legacy record"""
+    draft = service.create(identity, record)
+    record = service.publish(identity, draft.id)
+
+    # modify record in the db
+    db_record = RDMRecord.get_record(record._record.id)
+    db_record.parent.permission_flags = permission_flags
+    db_record.parent.commit()
+    db.session.commit()
+
+    return db_record
+
+
+def _add_to_community(record, community, service, db):
+    """Add a record to a community."""
+    record.parent.communities.add(community._record, default=False)
+    record.parent.commit()
+    record.commit()
+    db.session.commit()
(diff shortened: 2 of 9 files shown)


site/tests/conftest.py
```diff
@@ -8,17 +8,40 @@
 """Pytest fixtures."""
 
 from collections import namedtuple
+from copy import deepcopy
```
anikachurilova, line 11: Note: conftest to be rebased from https://github.com/zenodo/zenodo-rdm/pull/274

site/zenodo_rdm/legacy/requests/record_upgrade.py
```diff
@@ -0,0 +1,155 @@
...
+# Actions
+#
+class SubmitAction(actions.SubmitAction):
+    """Submit action."""
+
+    def execute(self, identity, uow):
+        """Execute the submit action."""
+        record = self.request.topic.resolve()
+        # Set the record's title as the request title.
+        self.request["title"] = record.metadata["title"]
```
ntarocco, line 49: We are missing to have an extra msg here, with a nice explanation of what is this request about.
Ping @slint for suggestions if he wants to have this in Zenodo.

site/zenodo_rdm/legacy/requests/record_upgrade.py
```diff
@@ -0,0 +1,155 @@
...
+
+class AcceptAction(actions.AcceptAction):
+    """Accept action."""
+
+    def execute(self, identity, uow):
+        """Grant access to restricted files of a record for a community."""
+        record = self.request.topic.resolve()
+
+        # remove flag from record parent, permissions logic will do the rest
+        _remove_permission_flag(record)
```
ntarocco, line 61: I would add an extra comment (event), stating something like:
`{user first name/last name} accepted the request. The community {name} will now have access to the restricted files.` or similar.

site/zenodo_rdm/legacy/requests/record_upgrade.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def execute(self, identity, uow):
+        """Deny access to restricted files of a record for all the communities."""
+        record = self.request.topic.resolve()
+
+        # remove the flag
+        _remove_permission_flag(record)
+
+        comment = {}
+        if record.access.protection.files == "public":
```
ntarocco, line 77:
```suggestion
        files_already_public = record.access.protection.files == "public"  # looks like the uploader changed the files restriction while the request is still open
        if files_already_public:
```

site/zenodo_rdm/legacy/requests/record_upgrade.py
```diff
@@ -0,0 +1,155 @@
...
+            comment = {
+                "payload": {
+                    "content": _(
+                        "The request has been automatically cancelled because the record doesn't have restricted files."
+                    )
+                }
+            }
+        elif not record.parent.communities.ids:
```
ntarocco, line 85:
```suggestion
        elif not record.parent.communities.ids:
            # it looks like the uploader removed the record from all communities while the request is still open
```

site/zenodo_rdm/legacy/requests/record_upgrade.py
```diff
@@ -0,0 +1,155 @@
...
+        if comment:
+            # add a comment and cancel the request
+            action_obj = actions.RequestActions.get_action(self.request, "cancel")
+            action_obj.execute(system_identity, uow)
+            current_events_service.create(
+                system_identity, self.request.id, comment, CommentEventType, uow=uow
+            )
+
+            # make event immediately available in search
+            uow.register(IndexRefreshOp(indexer=current_events_service.indexer))
```
ntarocco, line 103: do you also need to re-index the parent record?

site/zenodo_rdm/legacy/requests/record_upgrade.py
```diff
@@ -0,0 +1,155 @@
...
+        elif not record.parent.communities.ids:
+            comment = {
+                "payload": {
+                    "content": _(
+                        "The request has been automatically cancelled because the record doesn't belong to any community."
+                    )
+                }
+            }
+
+        if comment:
```
ntarocco, line 94: This is indeed enough, but IMO I would prefer to see something like:
```
cancel_request = files_already_public or not record.parent.communities.ids
if cancel_request:
```

site/zenodo_rdm/legacy/requests/record_upgrade.py
```diff
@@ -0,0 +1,155 @@
...
+    def execute(self, identity, uow):
+        """Deny access to restricted files of a record for all the communities."""
+        record = self.request.topic.resolve()
+
+        # remove the flag
+        _remove_permission_flag(record)
+
+        # remove record from communities only if files are restricted and
+        # record belongs to some communities
+        if (
```
ntarocco, line 122: This might be a problem. Imagine this use case:
1. the day of migration, the record with restricted files is in one community, so this request is opened.
2. later on, the uploader will add the record to another community (actually)

Issues:
- when the user adds the record to the 2nd community, (s)he will approve that the community's owner will have access to the files, but they will not: the flag will block that.
- when expiring, the record will be removed even from the 2nd community, which should probably not to be done.

site/zenodo_rdm/legacy/requests/utils.py
```diff
@@ -0,0 +1,75 @@
...
+                dsl.Q("term", **{"type": LegacyRecordUpgrade.type_id}),
+                dsl.Q("term", **{"is_open": True}),
+            ],
+        ),
+    )
+    return next(results.hits)["id"] if results.total > 0 else None
+
+
+@unit_of_work()
+def submit_record_upgrade_request(record, uow, comment=None):
```
ntarocco, line 39: This is handy, but it might be too slow when Zenodo will migrate data (one ES check per record). They might want to handle the sending of notifications (e-mails) differently.
@slint can you please comment on this?

site/zenodo_rdm/legacy/requests/utils.py
```diff
@@ -0,0 +1,75 @@
...
+    if exists(system_identity, record):
+        raise Exception(
+            "There is already an open legacy record upgrade request for this record."
+        )
+
+    type_ = current_request_type_registry.lookup(LegacyRecordUpgrade.type_id)
+    receiver = ResolverRegistry.resolve_entity_proxy(
+        {"user": record.parent.access.owned_by[0].owner_id}
+    ).resolve()
+    expires_at = datetime.now(tz=utc) + timedelta(days=365)
```
ntarocco, line 59:
```suggestion
    expires_at = datetime.utcnow() + timedelta(days=365)
```

site/tests/requests/test_record_upgrade_request.py
```diff
@@ -0,0 +1,584 @@
...
+    return response
+
+
+def _create_legacy_record(identity, record, permission_flags, db, service):
+    """Create and publish a legacy record"""
+    draft = service.create(identity, record)
+    record = service.publish(identity, draft.id)
+
+    # get a UUID of the record
+    record = RDMRecord.pid.resolve(record.id)
```
ntarocco, line 44: In the tests (not in the code), you can do:
```suggestion
    record = record._record
```

site/tests/requests/test_record_upgrade_request.py
```diff
@@ -0,0 +1,584 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 CERN.
+#
+# Zenodo is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Test Zenodo legacy record upgrade request."""
```
ntarocco, line 8: Fantastic work! Would I would add in the tests is to test that:
- when the request is opened, or declined, the upload can access the files, the community owner cannot.
- when the request is accepted, both can read files.

site/tests/conftest.py
ntarocco, line 11: It feels like we need RDM fixtures in pytest-invenio (or somewhere else), to avoid copy/pasting all this around.

ntarocco (commented):
Fantastic work! Good job! 👏🏼 I have added a few comments around ;)

site/zenodo_rdm/legacy/requests/record_upgrade.py
```diff
@@ -0,0 +1,155 @@
...
+from invenio_rdm_records.proxies import current_record_communities_service
+from invenio_records_resources.services.uow import IndexRefreshOp, RecordCommitOp
+from invenio_requests.customizations import CommentEventType, RequestType, actions
+from invenio_requests.proxies import current_events_service, current_requests_service
+
+
+def _remove_permission_flag(record):
+    """Remove can_community_read_files permission flag."""
+    try:
+        record.parent.permission_flags.pop("can_community_read_files")
```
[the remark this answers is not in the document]
anikachurilova: That will fail in case if `permission_flags` itself is None

site/zenodo_rdm/legacy/requests/record_upgrade.py
slint: Something like `Upgrade your legacy record {record.metadata.title}` would also be fine. I see that this is the "title" of the description below, so we can skip it below then.

site/zenodo_rdm/legacy/requests/record_upgrade.py
```diff
@@ -0,0 +1,182 @@
...
+            "<b>Upgrade legacy record request</b><br><br>"
+            "The current record uses legacy behaviours which is being removed from "
+            "the platform and is in need of being upgraded. Upgrading the record "
+            "will have the following impact:"
+            "<ul><li><b>Access: </b>curators of communities that this record "
+            "is part of will be granted view and edit access to your record (i.e. curators can see restricted "
+            "files and edit metadata)</li></ul>"
+            f"If the record has not been upgraded by <b>{expires_at}</b>, it will"
+            " be <u>removed from the communities</u> it is part of and upgraded."
+        )
```
slint, line 79:
```suggestion
        self.request["description"] = (
            "The current record uses legacy behavior which has been removed from "
            "the platform, and needs to be upgraded. Upgrading the record "
            "will have the following impact:"
            "<ul><li><b>Access: </b>curators of communities that this record "
            "is part of will be granted view and edit access to your record (i.e. curators can see restricted "
            "files and edit metadata)</li></ul>"
            f"If you do not upgrade the record by <b>{expires_at}</b>, it will "
            "be automatically <u>removed from its communities</u>."
        )
```

- Removed i18n (since we're not using translations in Zenodo)
- Typos/phrasing
- Changed wording from passive to active (to make it clear that the user has to do something :) )

site/zenodo_rdm/legacy/requests/record_upgrade.py
slint: Agree, that this would have unexpected behaviour for this edge case. Mitigating though seems like a much bigger hassle (e.g. comparing the date the record was added to the new community to the date of the legacy upgrade request). 

So bottom line, I would say we're fine with this as it is, as long as there's a comment/note here with this edge case.

slint (commented):
Thanks a lot, looks awesome and pretty well done!

Some comments on the texts wording and some edge cases, but otherwise LGTM.

slint (commented):
Thanks a lot, looks awesome and pretty well done!

Some comments on the texts wording and some edge cases, but otherwise LGTM.

site/zenodo_rdm/legacy/requests/record_upgrade.py
anikachurilova: @slint You mean only leave the title  `Upgrade your legacy record {record.metadata.title}` without any explanation in the description? 
The thing is, this description is the only place where we explain to the user what is happening any why they have to complete this request, so I would leave it.

site/tests/conftest.py
```diff
@@ -8,17 +8,40 @@
...
+from zenodo_rdm.legacy.requests.record_upgrade import LegacyRecordUpgrade
 
 
 @pytest.fixture(scope="module")
 def app_config(app_config):
     """Mimic an instance's configuration."""
+    supported_configurations = ["RDM_PERMISSION_POLICY"]
+
+    for config_key in supported_configurations:
+        app_config[config_key] = getattr(config, config_key, None)
```
alejandromumo, line 40: I would set each config manually here. It's easier to maintain and understand what's being used for tests and what is not.

site/tests/conftest.py
```diff
@@ -8,17 +8,40 @@
...
+
+    app_config["REQUESTS_REGISTERED_TYPES"] = [LegacyRecordUpgrade()]
+    app_config["REST_CSRF_ENABLED"] = False
```
alejandromumo, line 43: good finding !
I was having an error in my PR related to this and it was driving me a bit crazy 🤯

site/tests/conftest.py
```diff
@@ -82,5 +111,341 @@ def client_with_login(client, users):
     """Log in a user to the client."""
     user = users[0]
     login_user(user)
-    login_user_via_session(client, email=user.email)
+    login_user_via_session(client, user)
     return client
+
+
+@pytest.fixture(scope="module")
+def resource_type_v(app):
```
alejandromumo, line 119: Many of these will conflict with https://github.com/zenodo/zenodo-rdm/pull/352 but we'll fix it when the time comes

site/tests/conftest.py
```diff
@@ -82,5 +111,341 @@ def client_with_login(client, users):
...
+
+
+@pytest.fixture(scope="session")
+def UserFixture():
+    """Class to create user fixtures from."""
+    return UserFixture_
+
+
+@pytest.fixture()
+def minimal_record():
```
alejandromumo, line 232: minor: does it make sense to move some of these fixtures to a more specific directory? E.g. this `conftest.py` is the higher one in the hierarchy, meaning it contains fixtures that are relevant for every test sub-module (e.g. it sets up a testing app, client, etc). On the other hand, some of these fixtures are only used to test `requests` and maybe we can move it. Anyway, they can be moved up afterwards if needed. 
I'm asking this so maybe we can contain this conftest of growing indefinitely and avoid it becoming a huge monster.

alejandromumo (commented):
Tests are failing, are they passing locally?

site/tests/conftest.py
anikachurilova: You're right! However, this should be rebased from the permissions PR first (see here a better implementation of this https://github.com/zenodo/zenodo-rdm/pull/274/files#diff-cb157e965f2f4ebc88eaed8a8d0e5df39aa1552e8b7414ee396b35e44234a6c3R33)
Just in this PR, the `zenodo_rdm.permissions` doesn't exist yet

Issue #305, opened 2023-05-08
migration: record PIDs


slint:
- [x] DOI 
- [x] Concept DOI
- [x] OAI ID
- [x] Recid
- [x] Parent Recid

### Migration use cases

- Published external DOI
    - Without being edited
    - With deposit being edited and changed external DOI value
- Legacy records that have a DOI, but not a Concept DOI
- Records with edge cases DOIs, see [edge cases](https://github.com/zenodo/zenodo/blob/b4a4ab88238619811488777f6a2cc92b5d490a94/zenodo/modules/records/config.py#L115)

PR #319, opened 2023-05-09
migration: pids changes

zzacharo:


site/zenodo_rdm/migrator/transform/entries/records/records.py
```diff
@@ -44,18 +46,27 @@ def _recid(self, entry):
...
-                "provider": "datacite",
-                "identifier": record["doi"],
-            }
+            _doi = record["doi"]
+            if _doi.startswith(ZENODO_DATACITE_PREFIX):
+                r["doi"] = {
+                    "client": "datacite",
+                    "provider": "datacite",
+                    "identifier": record["doi"],
+                }
```
slint, line 63: shelve: there are some [edge cases](https://github.com/zenodo/zenodo/blob/b4a4ab88238619811488777f6a2cc92b5d490a94/zenodo/modules/records/config.py#L115) that we need to take into account.

zzacharo: logged it here https://github.com/zenodo/zenodo-rdm/issues/305

PR #324, opened 2023-05-10
drafts: skip grants with id

zzacharo:


slint (approved):
It would be to log them somehow, but I know it's not easy, so let's just open an issue to address later.

Issue #187, opened 2023-01-20
legacy: implement fields that require porting in InvenioRDM


slint:
> NOTE: This can be split up into smaller issues as well to implement groups of fields as we progress through milestones

- [ ] `access_conditions`
- [x] `notes`
- [x] `references`
- [ ] `communities`
- [x] `journal_title`
- [x] `journal_volume`
- [x] `journal_issue`
- [x] `journal_pages`
- [x] `conference_title`
- [x] `conference_acronym`
- [x] `conference_dates`
- [x] `conference_place`
- [x] `conference_url`
- [x] `conference_session`
- [x] `conference_session_part`
- [x] `imprint_publisher`
- [x] `imprint_isbn`
- [x] `imprint_place`
- [x] `partof_title`
- [x] `partof_pages`
- [x] `thesis_supervisors`
- [x] `thesis_university`
- [x] `locations`
- [x] `dates`
- [x] `languages`
- [x] `method`
- [x] `related_identifiers`
- [x] `version`   
- [x] `access_rights`
- [x] `embargo_date`

alejandromumo:
Missing fields:


```
- access_right     (doable now, however RDM has the format of 'record', 'files' and 'embargo' and each one 
- embargo_date     (only applies to embargoed)
- access_conditions     (only applies to restricted access records)
- doi    
- prereserve_doi   
- communities     (RDM has parent.communities or parent.review.communities)
- subjects    (needs a strategy to be defined first)
```

zzacharo:
`communities` should be probably be coordinated with `draft.communities` during migration so they share the same logic of either creating an inclusion request i.e `parent.review` or marking them as part of the draft i.e storing them in `parent.communities`.

slint:
`communities` field support split to https://github.com/zenodo/zenodo-rdm/issues/346

PR #330, opened 2023-05-12
legacy: added missing fields 

alejandromumo:
closes https://github.com/zenodo/zenodo-rdm/issues/187

### DIFF
--- a/site/zenodo_rdm/legacy/deserializers/schemas.py
+++ b/site/zenodo_rdm/legacy/deserializers/schemas.py
@@ -7,12 +7,22 @@
 
 """Zenodo legacy deserializer schemas."""
 
+from copy import deepcopy
 from datetime import date
 
+import pycountry
+from flask import current_app
 from invenio_records.dictutils import clear_none
+from marshmallow_utils.schemas import IdentifierSchema
 from nameparser import HumanName
+from werkzeug.local import LocalProxy
 
 from ..vocabularies.licenses import LEGACY_LICENSES, legacy_to_rdm
+from ..vocabularies.utils import _load_json
+
+record_identifiers_schemes = LocalProxy(
+    lambda: current_app.config["RDM_RECORDS_IDENTIFIERS_SCHEMES"]
+)
 
 # based on
 # DOI to id https://digital-repositories.web.cern.ch/zenodo/support/operations/#openaire-grants-import
@@ -238,13 +248,236 @@ class LegacyRecordTransform:
         """Transform notes of a legacy record."""
         if not data:
             return None
+        rdm_additional_descriptions = []
 
-        rdm_additional_descriptions = [{"description": data, "type": {"id": "other"}}]
+        notes = data.get("notes")
+        if notes:
+            rdm_additional_descriptions.append(
+                {"description": notes, "type": {"id": "other"}}
+            )
+
+        method = data.get("method")
+        if method:
+            rdm_additional_descriptions.append(
+                {"description": method, "type": {"id": "methods"}}
+            )
 
         return rdm_additional_descriptions
 
+    def _locations(self, data):
+        """Transform locations of a legacy record."""
+        if not data:
+            return None
+
+        features = []
+        for legacy_location in data:
+            lat = legacy_location.get("lat")
+            lon = legacy_location.get("lon")
+            place = legacy_location["place"]
+            description = legacy_location.get("description")
+
+            feature = {"place": place}
+
+            if lat and lon:
+                geometry = {"type": "Point", "coordinates": [lon, lat]}
+                feature.update({"geometry": geometry})
+
+            if description:
+                feature.update({"description": description})
+
+            features.append(feature)
+
+        return {"features": features}
+
+    def _subjects(self, data):
+        """Transform subjects of a legacy record.
+
+        RDM subjects translate to either legacy keywords or subjects.
+        """
+
+        def _from_keywords(keywords):
+            """Legacy keywords are free text strings.
+
+            They map to custom subjects.
+            """
+            return [{"subject": kw} for kw in keywords]
+
+        def _from_subjects(data):
+            """Maps RDM subjects to legacy subjects.
+
+            Legacy subjects are custom vocabularies.
+            """
+            # TODO we still did not define a strategy to map legacy subjects to rdm.
+            return []
+
+        keywords = data.get("keywords", [])
+        subjects = data.get("subjects", [])
+
+        if not (keywords or subjects):
+            return None
+
+        rdm_subjects = _from_keywords(keywords) + _from_subjects(subjects)
+
+        return rdm_subjects
+
+    def _dates(self, data):
+        """Transform dates of a legacy record."""
+        if not data:
+            return None
+
+        rdm_dates = []
+
+        for legacy_date in data:
+            start_date = legacy_date.get("start")
+            end_date = legacy_date.get("end")
+            description = legacy_date.get("description")
+            rdm_date = None
+            if start_date and end_date:
+                rdm_date = f"{start_date}/{end_date}"
+            else:
+                # RDM implements EDTF Level 0. Open intervals are not allowed
+                # TODO throw an error.
+                continue
+
+            rdm_date = {
+                "date": rdm_date,
+                # Type is required on legacy Zenodo
+                "type": {
+                    # Type is a vocabulary on ZenodoRDM.
+                    "id": legacy_date["type"].lower(),
+                },
+            }
+
+            if description:
+                rdm_date.update({"description": description})
+            rdm_dates.append(rdm_date)
+
+        return rdm_dates
+
+    def _references(self, data):
+        """Transform references of a legacy record."""
+        if not data:
+            return None
+
+        rdm_references = []
+
+        for legacy_reference in data:
+            _rdm_reference = {"reference": legacy_reference}
+            rdm_references.append(_rdm_reference)
+
+        return rdm_references
+
+    def _language(self, data):
+        """Transform language of a legacy record."""
+
+        def _map_iso_639_1_to_639_2(language):
+            """Maps ISO 639-1 (2 digits) to ISO 639-2 (3 digits).
+
+            RDM implements 639-3 (3 digits), which is an extension of 639-2.
+            Therefore, mapping from 639-1 to 639-2 should be enough.
+
+            See: https://www.loc.gov/standards/iso639-2/php/code_list.php
+            """
+            mapping = pycountry.languages.get(alpha_2=language)
+            if mapping:
+                return mapping.alpha_3
+
+            return None
+
+        if not data:
+            return None
+
+        language = None
+
+        # Case 1 - 3 letter code
+        if len(data) == 3:
+            language = data
+        # Case 2  - 2 letter code
+        elif len(data) == 2:
+            language = _map_iso_639_1_to_639_2(data)
+
+        return [{"id": language}] if language else None
+
+    def _related_identifiers(self, data):
+        """Transform related identifiers of a legacy record."""
+        if not data:
+            return None
+
+        related_identifiers = []
+        identifier_schema = IdentifierSchema(allowed_schemes=record_identifiers_schemes)
+        for legacy_identifier in data:
+            # Identifier schema is used to detect the identifier's 'scheme'.
+            # In legacy, 'scheme' is not passed as a parameter. Instead, it's detected from the identifier itself.
+            # TODO schema.load() might error (e.g. invalid identifier scheme).
+            # TODO question: should fields error? They're "ignoring" errors for now (E.g. returning 'None')
+
+            # Patches a typo legacy relation type (see https://github.com/zenodo/zenodo/issues/1366)
+            if legacy_identifier["relation"] == "isOrignialFormOf":
+                legacy_identifier["relation"] = "isOriginalFormOf"
+
+            identifier = identifier_schema.load(
+                {
+                    "identifier": legacy_identifier["identifier"],
+                }
+            )
+            rdm_identifier = {
+                **identifier,
+                "relation_type": {
+                    # relation_type is a vocabulary
+                    "id": legacy_identifier["relation"].lower(),
+                },
+            }
+
+            # Resource type is optional.
+            legacy_type = legacy_identifier.get("resource_type")
+            if legacy_type:
+                rdm_identifier.update({"resource_type": {"id": legacy_type}})
+
+            related_identifiers.append(rdm_identifier)
+
+        return related_identifiers
+
+    def _alternate_identifiers(self, data):
+        """Transform alternate identifiers of a legacy record."""
+        if not data:
+            return None
+
+        alternate_identifiers = []
+        identifier_schema = IdentifierSchema(allowed_schemes=record_identifiers_schemes)
+
+        for legacy_identifier in data:
+            rdm_identifier = identifier_schema.load(
+                {
+                    "identifier": legacy_identifier["identifier"],
+                }
+            )
+
+            alternate_identifiers.append(rdm_identifier)
+
+        return alternate_identifiers
+
     def _metadata(self, data):
         """Transform the metadata of a record."""
+
+        def _pre_load(data):
+            # Some identifiers are alternate identifiers if the relation is "isAlternateIdentifier"
+            input_identifiers = data.get("related_identifiers", [])
+            alternate_identifiers = []
+            related_identifiers = []
+            for identifier in input_identifiers:
+                if identifier.get("relation", "") == "isAlternateIdentifier":
+                    alternate_identifiers.append(identifier)
+                else:
+                    related_identifiers.append(identifier)
+
+            data["related_identifiers"] = related_identifiers
+            data["alternate_identifiers"] = alternate_identifiers
+
+            return data
+
+        original = deepcopy(data)
+        data = _pre_load(original)
+
         if data:
             return {
                 "title": data.get("title"),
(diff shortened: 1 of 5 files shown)


site/zenodo_rdm/legacy/serializers/schemas.py
```diff
@@ -105,6 +105,46 @@ class ThesisSchema(Schema):
...
+            start = None
+            end = None
+            # EDTF level 0 specifies intervals using "/" (e.g. 2004-02-01/2005-02)
+            if is_interval:
+                start = interval[0]
+                end = interval[1]
+                result["start"] = start
+                result["end"] = end
+            else:
+                # TODO what to do here? We do not know whether the date is start, end or exact date (start == end)
```
alejandromumo, line 134: This case is particularly troublesome. 

In RDM, there is no concept of `start` or `end` when a date is a single date (e.g. `2023-02-02`). How do we convert it back to legacy?

[EDTF](https://www.loc.gov/standards/datetime/) implements this in level 1, however RDM implements EDTF level 0 for now.

site/zenodo_rdm/legacy/serializers/schemas.py
```diff
@@ -163,6 +215,40 @@ def dump_resource_type(self, result, original, **kwargs):
...
+        subjects = original.get("subjects", [])
+        serialized_subjects = []
+        serialized_keywords = []
+        if subjects:
+            for _sbj in subjects:
+                _id = _sbj.get("id")
+                _subject = _sbj.get("subject")
+                # If subject has an id, it's a controlled vocabulary
+                if _id:
+                    # TODO we still did not define a strategy to map legacy subjects to rdm.
```
alejandromumo, line 230: Still needs to be done for legacy and migration

site/zenodo_rdm/legacy/deserializers/schemas.py
```diff
@@ -238,11 +248,203 @@ def _additional_descriptions(self, data):
...
+        """Transform related identifiers of a legacy record."""
+        if not data:
+            return None
+
+        related_identifiers = []
+        identifier_schema = IdentifierSchema(allowed_schemes=record_identifiers_schemes)
+        for legacy_identifier in data:
+            # Identifier schema is used to detect the identifier's 'scheme'.
+            # In legacy, 'scheme' is not passed as a parameter. Instead, it's detected from the identifier itself.
+            # TODO schema.load() might error (e.g. invalid identifier scheme). Other fields are not erroring.
```
alejandromumo, line 424: error handling is going to be tackled in a separate [issue](https://github.com/zenodo/zenodo-rdm/issues/334)

site/zenodo_rdm/legacy/serializers/schemas.py
```diff
@@ -163,6 +219,64 @@ def dump_resource_type(self, result, original, **kwargs):
...
+
+        legacy_access = None
+
+        if is_open:
+            legacy_access = "open"
+        elif is_restricted:
+            legacy_access = "restricted"
+        elif is_embargoed:
+            legacy_access = "embargoed"
+        # TODO how to map to closed access?
```
alejandromumo, line 272: Currently we can deserialize access `closed` (legacy) to `files->restricted` (RDM).
However we are not able to serialize it back.

site/zenodo_rdm/legacy/serializers/schemas.py
```diff
@@ -163,6 +219,64 @@ def dump_resource_type(self, result, original, **kwargs):
...
+
+        if not legacy_access:
+            # Throw an error maybe?
```
alejandromumo, line 275: theoretically every access should be mapped. 
I did not add any exceptions for it to be tackled in a separate issue

site/zenodo_rdm/legacy/deserializers/schemas.py
```diff
@@ -261,10 +463,76 @@ def _metadata(self, data):
...
+            }
+
+        default = _open_record()
+        access = None
+        metadata = data.get("metadata", {})
+        access_right = metadata.get("access_right")
+        embargo_date = metadata.get("embargo_date")
+
+        if not access_right:
+            # TODO throw an error actually? Or default to public?
```
alejandromumo, line 514: This is more of a validation check. Access right should be there always, so I think should be tackled in a separate [issue](https://github.com/zenodo/zenodo-rdm/issues/334) when we defined a strategy for error handling.

site/zenodo_rdm/legacy/deserializers/schemas.py
```diff
@@ -261,10 +463,76 @@ def _metadata(self, data):
...
+        # TODO closed access maps to open?
+        if is_open:
+            access = _open_record()
+        elif is_restricted or is_closed:
+            access = _restricted_record()
+        elif is_embargoed and embargo_date:
+            access = _embargoed_record(embargo_date)
+
+        if not access:
+            # TODO throw an error.
```
alejandromumo, line 531: This should not happen anytime so can be removed probably.

alejandromumo:
I added communities on a separate commit as it's bypassing the service's public interface. 

Right now, this will generate 1 inclusion request per community that must be accepted afterwards.

**NOTE** : it doesn't work if a community has direct publish enabled!!! It will fail in the DB because two entries are added in `parent_records_metadata` with the same pair of (`community_id`, `record_id`)

site/zenodo_rdm/legacy/services.py
```diff
@@ -84,15 +96,60 @@ class LegacyRecordService(RDMRecordService):
...
+        """Creates community inclusion requests for a record's communities.
+
+        This bypasses the record communities service checks on community add.
+        """
+        if not communities:
+            return
+
+        processed = []
+        errored = []
+        for community in communities:
```
alejandromumo, line 133: This code was taken from [rdm-records](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/communities/service.py#L129). 

The idea was to bypass the [pid resolve](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/communities/service.py#LL125C26-L125C26) because it does not exist yet (the draft is not published)

site/zenodo_rdm/legacy/services.py
```diff
@@ -84,15 +96,60 @@ class LegacyRecordService(RDMRecordService):
...
+            res = self.pids.create(
+                identity=identity,
+                id_=res.id,
+                scheme="doi",
+                expand=True,
+                uow=uow,
+            )
+
+        communities = data.get("communities", [])
+        p = self.add_communities(communities, identity, uow, res._record)
```
alejandromumo, line 114: question: if it fails while adding communities, should we rollback all of it or ignore?

site/zenodo_rdm/legacy/serializers/schemas.py
```diff
@@ -321,8 +496,16 @@ def dump_state(self, result, original, **kwargs):
     @post_dump(pass_original=True)
     def dump_prereserve_doi(self, result, original, **kwargs):
         """Dump prereserved DOI information."""
+        provider = original["pids"]["doi"]["provider"]
+        recid = original["id"]
+        # For external DOIs, the prereserve_doi is hacked on the response.
+        # TODO confirm this  hack
```
alejandromumo, line 502: For external dois, the `prereserve_doi` is not defined. Therefore I added it here manually

site/zenodo_rdm/legacy/serializers/schemas.py
slint: See my comment on the deserializer, TLDR: either make sure we have always both start/end in legacy or check how easy it is to use EDTF Level 1 in RDM.

site/zenodo_rdm/legacy/serializers/schemas.py
slint: minor: actually that's `closed` access in legacy, since `restricted` is a special case of `closed` for which we need the "Access requests" feature to be implemented).

site/zenodo_rdm/legacy/serializers/schemas.py
```diff
@@ -321,8 +496,16 @@ def dump_state(self, result, original, **kwargs):
...
+        if provider == "external":
+            doi = f"10.5281/{recid}"
```
slint, line 504: minor: should be `doi = f"10.5281/zenodo.{recid}"` [as in legacy](https://github.com/zenodo/zenodo/blob/b4a4ab88238619811488777f6a2cc92b5d490a94/zenodo/modules/records/minters.py#LL43C26-L43C26).

site/zenodo_rdm/legacy/services.py
```diff
@@ -84,13 +95,20 @@ class LegacyRecordService(RDMRecordService):
...
-            id_=res.id,
-            scheme="doi",
-            expand=True,
-            uow=uow,
-        )
+        pids = data.get("pids", {})
+
+        # If pids is provided, it is created automatically.
+        # TODO confirm this is correct
+        # TODO what happens with other providers / schemes? They are allowed in RDM
```
slint, line 103: Hopefully we never get from the deserializer other types of PIDs since we control it :)

slint (commented):
*(Review of metadata only, i.e. not `communities`)*

Nice and very thorough work! Some minor comments and possible improvements, and some things to shelve or address separately.

site/zenodo_rdm/legacy/services.py
```diff
@@ -109,8 +110,46 @@ def create(self, identity, data, uow=None, expand=False):
                 uow=uow,
             )
 
+        communities = data.get("communities", [])
```
slint, line 113: minor: shouldn't we actually pop this before passing it to `super().create(...)` above?

site/zenodo_rdm/legacy/services.py
slint: We should be consistent with how it worked in legacy, in which case we're never failing but just ignore/skip (in case the request is already created or similar). An alternative is to handle this case here inside the service method and check before making the `add_communities` call (so that we only add the new ones). Basically do "look before you leap" instead of the usual "ask for forgiveness than permission".

In the same fashion, we should also handle removing communities :)

site/zenodo_rdm/legacy/services.py
slint: I think we, discussed something similar in the chat, but the idea would be to see if we could convert this minimal prototype into a couple of flags or config that could make this into a feature (or at least possible via normal non-private service calls).

slint (commented):
*(Review of the communities part)*

Some very basic comments, but I guess there's a lot more to make this working. We should probably break this PR into two, since communities support will require a lot of work that's unrelated to the missing metadata fields.

site/zenodo_rdm/legacy/deserializers/schemas.py
```diff
@@ -9,10 +9,18 @@
...
+from flask import current_app
 from invenio_records.dictutils import clear_none
+from marshmallow_utils.schemas import IdentifierSchema
 from nameparser import HumanName
+from werkzeug.local import LocalProxy
 
 from ..vocabularies.licenses import LEGACY_LICENSES, legacy_to_rdm
+from ..vocabularies.utils import _load_json
+
+record_identifiers_schemes = LocalProxy(
```
ppanero, line 21: usually this kind of variables go to the config.py, avoiding LocalPorxy here might make it cleaner.

site/zenodo_rdm/legacy/deserializers/schemas.py
```diff
@@ -238,11 +248,203 @@ def _additional_descriptions(self, data):
...
+        keywords = data.get("keywords", [])
+        subjects = data.get("subjects", [])
+
+        if not (keywords or subjects):
+            return None
+
+        rdm_subjects = []
+
+        kws = _subjects_from_keywords(keywords)
+        if kws:
```
ppanero, line 322: if ` _subjects_from_keywords` and `_subjects_from_subjects` return [] when empty you can avoid ifs and have simply:

```python
rdm_subjects = _subjects_from_keywords(keywords) + _subjects_from_subjects(keywords)
```

site/zenodo_rdm/legacy/deserializers/schemas.py
```diff
@@ -238,11 +248,203 @@ def _additional_descriptions(self, data):
...
+        """
+
+        def _subjects_from_keywords(keywords):
+            """Legacy keywords are free text strings.
+
+            They map to custom subjects.
+            """
+            return [{"subject": kw} for kw in keywords]
+
+        def _subjects_from_subjects(data):
```
ppanero, line 305: naming: a bit redundant, since they are inside `_subjects` what about `_from_keywords` and `_from_subjects`

site/zenodo_rdm/legacy/deserializers/schemas.py
```diff
@@ -238,11 +248,203 @@ def _additional_descriptions(self, data):
...
+        for legacy_reference in data:
+            _rdm_reference = {"reference": legacy_reference}
+            rdm_references.append(_rdm_reference)
+
+        return rdm_references
+
+    def _language(self, data):
+        """Transform language of a legacy record."""
+
+        def _map_iso_639_1_to_639_2(language):
```
ppanero, line 386: invenio already installs `pycountry`, what about using that one?

e.g.

```python
>>> import pycountry
>>> pycountry.languages.get(alpha_2="de")
Language(alpha_2='de', alpha_3='deu', bibliographic='ger', name='German', scope='I', type='L')
>>> pycountry.languages.get(alpha_2="de").alpha_3
'deu'
```

site/zenodo_rdm/legacy/deserializers/schemas.py
ppanero: Wouldnt it be safer to translate to closed? to avoid data leaks of datasets? i.e. user makes a mistake and sends unintendedly open, but should have been closed.

I guess is our "part" on promoting open data, the user can always go an edit. Right?

site/zenodo_rdm/legacy/serializers/schemas.py
```diff
@@ -145,11 +202,11 @@ def hook_imprint_partof(self, data, **kwargs):
         publisher = data.get("publisher")
         if imprint:
             data["part_of"] = imprint
-        if publisher:
-            data["custom_fields"]["imprint:imprint"] = {
-                **imprint,
-                "publisher": publisher,
-            }
+            if publisher:
```
ppanero, line 205: is this indent intended? (tongue twister haha)

alejandromumo: yes, the RDM custom field `imprint:imprint` serializes to two fields in legacy (`imprint` and `part_of`). 

If only `imprint` is set, we can hook it to `part_of` (serializes to `partof_title` and `partof_pages`).

On the other hand if there's also a publisher, we add it to the imprint information (that serializes to `imprint_publisher`, `imprint_isbn` and `imprint_place`).

site/zenodo_rdm/legacy/deserializers/schemas.py
```diff
@@ -238,11 +248,203 @@ def _additional_descriptions(self, data):
...
+                if start_date == end_date:
+                    rdm_date = start_date
+                else:
+                    rdm_date = f"{start_date}/{end_date}"
+            elif start_date:
+                rdm_date = start_date
+
+            elif end_date:
+                rdm_date = end_date
+            # No start_date nor end_date is not allowed on legacy Zenodo
```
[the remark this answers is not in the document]
alejandromumo: Maybe I missunderstood your comment but I think legacy requires at least one date to be set:

```python
if not start and not end:
    raise ValidationError()
```

I added the comment in the code to explain that no `else` clause was added because legacy does not allow that scenario (no start and no end date). In fact, in this case the legacy api should also return an error.

Open intervals are another thing. Our problem with open intervals is  that we can't serialize an RDM date to a legacy date when the interval is open (and for this we need EDTF Level 1, as you said.)

site/zenodo_rdm/legacy/deserializers/schemas.py
alejandromumo: I think both have good points.

Also, the same can happen when uploading through the UI (e.g. user forgets about the access rights). In that case, the UI defaults to open access. 

About mistakes: yeah that would be a pain for the user, especially if they're uploading a bunch of records using a script. I think they would have to change it in the UI, one by one.

site/zenodo_rdm/legacy/vocabularies/data/languages_iso_639_1_to_2_mapping.json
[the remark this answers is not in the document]
alejandromumo: yes, also suggested by @ppanero and definitely the way to go! I was not aware of it, thanks guys!

site/zenodo_rdm/legacy/serializers/schemas.py
alejandromumo: moved it [here](https://github.com/zenodo/zenodo-rdm/pull/347)

site/zenodo_rdm/legacy/serializers/schemas.py
alejandromumo: I am not sure what you meant by that, sorry 😅  
For now I dropped a TODO here to signal that we need to come back to this after access requests are implemented

site/zenodo_rdm/legacy/deserializers/schemas.py
```diff
@@ -261,10 +463,87 @@ def _metadata(self, data):
...
+            pass
+
+        return access
+
+    def _pids(self, data):
+        """Transform legacy doi in RDM pid with external provider."""
+        metadata = data.get("metadata")
+        doi = metadata.get("doi")
+
+        if not doi:
```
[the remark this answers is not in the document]
alejandromumo: I am not sure what happens here.

If the user provides a prereserved DOI, in RDM it should be seen as an external or internal pid?  In case of external, who's the provider? Datacite?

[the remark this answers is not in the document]
alejandromumo: The only difference would be the provider, right? E.g. if the passed doi is a zenodo doi, the the provider is datacite.

site/zenodo_rdm/legacy/deserializers/schemas.py
alejandromumo: in this case `site/zenodo_rdm/config.py`?

site/zenodo_rdm/legacy/deserializers/schemas.py
```diff
@@ -238,11 +248,203 @@ def _additional_descriptions(self, data):
         """Transform notes of a legacy record."""
         if not data:
             return None
+        rdm_additional_descriptions = []
+
+        notes = data.get("notes")
+        if notes:
+            rdm_additional_descriptions.append(
+                {"description": notes, "type": {"id": "other"}}
```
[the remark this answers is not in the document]
slint: https://github.com/zenodo/zenodo-rdm/issues/349

site/zenodo_rdm/legacy/deserializers/schemas.py
```diff
@@ -238,11 +248,203 @@ def _additional_descriptions(self, data):
...
+        # Case 1 - 3 letter code
+        if len(data) == 3:
+            language = data
+        # Case 2  - 2 letter code
+        elif len(data) == 2:
+            language = _map_iso_639_1_to_639_2(data)
+
+        return [{"id": language}] if language else None
+
+    def _related_identifiers(self, data):
```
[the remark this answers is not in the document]
slint: https://github.com/zenodo/zenodo-rdm/issues/350

site/zenodo_rdm/legacy/deserializers/schemas.py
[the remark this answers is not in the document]
slint: Exactly, we need to translate it to the correct provider value, i.e. `datacite`. Though I think this has to happen in the service-method, since it's a separate call to the `records_service.pids` sub-service.

slint (commented):
Nicely done! One minor comment on where to deal the external/datacite provider PIDs.

I've also opened some new issues for the shelve comments.

site/zenodo_rdm/legacy/deserializers/schemas.py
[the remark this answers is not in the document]
alejandromumo: You mean that the provider should be returned by the pid manager based on the identifier? E.g. [here](https://github.com/inveniosoftware/invenio-rdm-records/blob/6d7758714d08be9317bed5d402fc09731f71bf3d/invenio_rdm_records/services/pids/manager.py#L29)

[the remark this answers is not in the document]
slint: The issue is that the behaviour of passing the `pids` field as input via the `create()/update()` service methods is a bit flaky (even in the "stock" RDM API). E.g. if one passes a `doi` with `datacite` provider, and then changes over to `external`, there are additional calls necessary to un-reserve the `datacite` DOI... You can check the REST API calls in the default InvenioRDM deposit form and see that there are more calls happening to `/api/records/{id}/draft/pids`. These service calls must be "baked-in" the Zenodo legacy REST API.

Since this is actually a bit difficult to handle fully now, I would say let's make the happy paths work (i.e. a) publish with Zenodo DOI, and b) publish with an external DOI), and we tackle the edge cases separately.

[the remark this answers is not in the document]
slint: https://github.com/zenodo/zenodo-rdm/issues/353

slint (approved):

PR #351, opened 2023-05-22
Awards/funders loading

### LINKED ISSUE #323: migration: funders + awards dump loading

slint:
Closes #323.

- migrator: fix users query for roles cache
- config: publish kwargs for bulk indexer retries
- migrator: add awards/funders dumps and loading

### DIFF
--- a/site/zenodo_rdm/migrator/scripts/dump_awards_db.py
+++ b/site/zenodo_rdm/migrator/scripts/dump_awards_db.py
@@ -1,0 +1,126 @@
+"""Parse OpenAIRE awards dumps into CSV format, importable via COPY.
+
+A modified version of https://github.com/inveniosoftware/invenio-vocabularies/blob/master/invenio_vocabularies/contrib/awards/datastreams.py
+for the purpose of producing an easy to load CSV dump of the awards into an InvenioRDM instance.
+
+To use call ``load_files(DATA_PATHS, "awards.csv")``.
+"""
+import csv
+import gzip
+import json
+import uuid
+
+from invenio_rdm_migrator.utils import ts
+
+DATA_PATHS = [
+    "awards-apr-2023.jsonl.gz",  # https://zenodo.org/record/7870151
+    "awards-mar-2023.jsonl.gz",  # https://zenodo.org/record/7803150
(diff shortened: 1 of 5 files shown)


site/zenodo_rdm/migrator/scripts/dump_awards_db.py
ppanero, line 1: We have a [yaml writer](https://github.com/inveniosoftware/invenio-vocabularies/blob/master/invenio_vocabularies/datastreams/writers.py#L89), could implement a csv writter and then we could use datastreams to do this processing.

site/zenodo_rdm/migrator/scripts/dump_awards_db.py
```diff
@@ -0,0 +1,131 @@
...
+    "snsf________": "00yjd3n13",
+    "sshrc_______": "006cvnv84",
+    "tubitakf____": "04w9kkr77",
+    "ukri________": "001aqnf71",
+    "wt__________": "029chgv08",
+}
+VOCABULARIES_AWARDS_EC_ROR_ID = "00k4n6c32"
+
+
+def _ts():
```
ppanero, line 53: this is available in invenio-rdm-migrator (can be imported) https://github.com/inveniosoftware/invenio-rdm-migrator/blob/master/invenio_rdm_migrator/utils.py

ppanero (commented):
LGTM and can be merged. However, I wonder if we could have them as streams in the same fashion we have record files. To avoid splitting logic (load function seems quite the same if not the same) and reduce manual steps. Maybe this could be tackled as a separate issue.

site/zenodo_rdm/migrator/scripts/dump_awards_db.py
slint: Agree; there are two options I see for integrating:

- In the regular migrator streams via the `existing_data: true` flag (probably we'll go with this one)
- Add a new `invenio-vocabularies` writer:
  - For dumping the CSV
  - For directly running `COPY ...` with the output

ppanero: I see them as complementary:
- the second one could help create the csv files
- the first one load them in the migration process

Issue #299, opened 2023-05-05
legacy api: add tests for serializer and deserializer 


alejandromumo:
relates to : https://github.com/zenodo/zenodo-rdm/pull/289

PR #352, opened 2023-05-22
legacy: added api tests

alejandromumo:
closes https://github.com/zenodo/zenodo-rdm/issues/299

- Added legacy tests (ported from   [legacy](https://github.com/zenodo/zenodo/blob/master/tests/unit/deposi/test_api_metadata.py)).
- Refactored deserializer to be a marshmallow schema. 
  - With tests, we needed a way of adding validaton to the schema so we decided to refactor it now.


site/tests/legacy/test_deposit_api.py
```diff
@@ -0,0 +1,316 @@
...
+    # doi is returned as a top level key (and not inside metadata)
+    assert data["doi"] == test_data["metadata"]["doi"]
+    ignored_keys.add("doi")
+
+    differences = list(
+        dictdiffer.diff(data["metadata"], expected_record_metadata, ignore=ignored_keys)
+    )
+
+    # Known issues:
+    # - related_identifiers relation expects 'isCitedBy' but returns lower cased.
```
alejandromumo, line 313: @slint these are the two failing scenarios:

- `isCitedBy` can be mapped back (or simply ignored)
- I think `additional_descriptions` might be failing due to a vocabulary missing, however this can easily be tested by running a local instance and executing the same request.

site/tests/legacy/test_deposit_api.py
```diff
@@ -0,0 +1,316 @@
...
+
+from zenodo_rdm.legacy.resources import LegacyRecordResourceConfig
+
+
+@pytest.fixture(scope="function")
+def test_data():
+    return dict(
+        metadata=dict(
+            access_right="embargoed",
+            # TODO uncomment when communities are implemented
```
alejandromumo, line 30: Or simply accept them and discard them in the serializer

site/zenodo_rdm/legacy/serializers/schemas.py
```diff
@@ -143,13 +149,28 @@ class RelatedIdentifierSchema(Schema):
...
+
+        # Or throw an error
+        return legacy_relation or missing
 
 
 class MetadataSchema(Schema):
     """Metadata schema."""
 
+    # TODO missing conference
+
```
slint, line 173: minor: Isn't it below actually?

site/zenodo_rdm/legacy/deserializers/schemas.py
```diff
@@ -7,168 +7,46 @@
...
+    #         "access": self._access(data),
+    #         "files": {"enabled": True},
+    #         "pids": self._pids(data),
+
+    metadata = fields.Nested(MetadataSchema)
+    files = fields.Constant({"enabled": True})
+    # Fields are added on post load
+    custom_fields = fields.List(fields.Dict())
+    access = fields.Dict()
+    pids = fields.Dict()
```
slint, line 35: minor: we can remove then, so that we don't accidentally accept them as input in the payload

site/zenodo_rdm/legacy/deserializers/metadata.py
```diff
@@ -0,0 +1,487 @@
...
+            data["alternate_identifiers"] = alternate_identifiers
+
+        return data
+
+    title = SanitizedUnicode()
+    description = SanitizedUnicode()
+    publication_date = SanitizedUnicode(default=date.today().isoformat())
+    resource_type = fields.Str()
+    creators = fields.List(fields.Nested(CreatorSchema))
+    publisher = fields.Constant("Zenodo")  # TODO is it?
```
slint, line 187: minor: Yes, on legacy we didn't have support for custom Publisher. `imprint_publisher` though could go here (since we don't have it in RDM, and rely on `publisher`)

slint (commented):
Nice! Some minor things, but otherwise looks good from a quick overview.

site/zenodo_rdm/legacy/serializers/schemas.py
alejandromumo: Yes, it was as leftover 😅 thanks!

Issue #348, opened 2023-05-22
migration, transformer: improve related identifiers parsing


ppanero:
The following cases (obtained from log files of a full run) have made the transformation step fail on the `related_identifiers` field:

```
'related_identifiers': [{
	'identifier': '978-65-997142-0-7'
}]


'related_identifiers': [{
	'scheme': 'ads',
	'relation': 'isIdenticalTo',
	'identifier': '2004PhDT........43B'
}, {
	'relation': 'isIdenticalTo',
	'identifier': 'tel-00011634'
}]


'alternate_identifiers': [{
	'identifier': 'local:'
}]


'related_identifiers': [{
	'relation': 'isSupplementTo',
	'identifier': ' arXiv:1601.08082'
}, {
	'relation': 'isSupplementTo',
	'identifier': 'hello'
}, {
	'relation': 'isSupplementTo',
	'identifier': 'arXiv:1601_.08082'
}, {
	'scheme': 'arxiv',
	'relation': 'isSupplementTo',
	'identifier': 'arXiv:1601.08082'
}]


'related_identifiers': [{
	'scheme': 'url',
	'relation': 'isSupplementTo',
	'identifier': 'https://github.com/genomematt/xenomapper/tree/v1.0.1'
}],
'alternate_identifiers': [{
	'identifier': '10.21105.joss.00018'
}]


'related_identifiers': [{
	'relation': 'isSupplementTo',
	'identifier': 'ISSN'
}]
```

PR #355, opened 2023-05-23
migrator, transform: ensure validation of identifiers

ppanero:
- closes https://github.com/zenodo/zenodo-rdm/issues/348

### DIFF
--- a/site/zenodo_rdm/migrator/transform/entries/records/metadata.py
+++ b/site/zenodo_rdm/migrator/transform/entries/records/metadata.py
@@ -9,11 +9,14 @@
 
 from urllib.parse import urlparse
 
+from idutils import detect_identifier_schemes
 from invenio_rdm_migrator.transform import Entry, drop_nones
 from nameparser import HumanName
 
 from zenodo_rdm.legacy.deserializers.schemas import FUNDER_DOI_TO_ROR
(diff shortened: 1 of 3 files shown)


site/zenodo_rdm/migrator/transform/entries/records/metadata.py
```diff
@@ -140,6 +143,18 @@ def _subjects(cls, keywords):
...
+    def _validate_identifier(cls, identifier):
+        """Validates the information of an identifier and completes it if needed."""
+        value = identifier.get("identifier", "")
+        scheme = identifier.get("scheme")
+        if not scheme:
+            guess = detect_identifier_schemes(value)
+            if guess:
+                identifier["scheme"] = guess[0]  # default to first guess
+            else:
+                raise InvalidIdentifier(identifier)
```
ppanero, line 156: This means the full record won't be migrated until we fix it. Leads me to question:
- For now we want to raise, so we can investigate and fix.
- Later we might want to let records pass and ignore this malformed values (i.e. remove), which might be the same action we do to fix them.

WDYT?

slint: Agree, the best would be just to log these errors and either skip these identifiers and get the record in or include them and deal with it later... I think the bigger task here is to figure out a way to manage logging in a better way.

slint (approved):

Issue #311, opened 2023-05-08
migration: deposit files


slint:

ppanero:
It is blocked because we need to install an extension in the DB (needs DBoD service intervention), but the code is reviewable

PR #358, opened 2023-05-25
Draft files

ppanero:
- requires https://github.com/inveniosoftware/invenio-rdm-migrator/pull/89
- closes https://github.com/zenodo/zenodo-rdm/issues/311

### DIFF
--- a/site/zenodo_rdm/migrator/extract.py
+++ b/site/zenodo_rdm/migrator/extract.py
@@ -231,60 +231,6 @@ COPY (
 ) TO STDOUT;
 """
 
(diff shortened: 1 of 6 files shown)


site/zenodo_rdm/migrator/extract.py
```diff
@@ -231,60 +231,6 @@ def run(self):
 ) TO STDOUT;
 """
 
-# TODO: Check if we just dump the entire tables (files, objects, buckets) and
```
ppanero, line 234: Removing it now, will be re-added (slightly modified) as a post_load hook for draft files

PR #361, opened 2023-05-25
Empty funders

zzacharo:


ppanero (approved):
Maybe add also awards.yaml? While there are no defaults in rdm-records for now, better safe than sorry?

slint (approved):

PR #365, opened 2023-05-26
Funders awards streams

zzacharo:


site/zenodo_rdm/migrator/load.py
```diff
@@ -33,7 +32,34 @@ def _generate_rows(self, **kwargs):
...
+    """Zenodo to RDM table generator for files."""
+
+    def __init__(self):
+        """Constructor."""
+        super().__init__(
+            tables=[FilesInstance, FilesBucket, FilesObjectVersion],
+        )
+
+
+class ZenodoFilesLoad(ZenodoExistingDataLoadBase):
```
ppanero, line 62: remove _validate method on this class?

ppanero (commented):
IMO there is no need to create all the classes, only `ZenodoExistingDataTableGenaratorBase` and `ZenodoExistingDataLoadBase` would suffice then you can simply instantiate, which would simplify the code a lot. If needed, merge and we can refactor later.

```py
awards = ZenodoExistingDataLoadBase(table_generators=[ZenodoExistingDataTableGenaratorBase(tables=[Awards])]
```

Issue #336, opened 2023-05-15
legacy: community about bage redirect not necessary


slint:
We have set up a redirection rule for `/communities/{id}/about -> /communities/{id}`, which is not needed anymore, since there is a dedicated about page now (see [example](https://inveniordm.web.cern.ch/communities/cattos/about)). It looks like we've gone full circle :)

PR #383, opened 2023-06-02
Removing obsolete redirection

psaiz:
This is to close https://github.com/inveniosoftware/invenio-app-rdm/issues/2217 and closes https://github.com/zenodo/zenodo-rdm/issues/336.

The redirection for communities/<community>/about doesn't seem to work. We could rely on the redirection from invenio-rdm, instead of having to redefine it in zenodo-rdm

### DIFF
--- a/site/zenodo_rdm/config.py
+++ b/site/zenodo_rdm/config.py
@@ -221,10 +221,6 @@ REDIRECTOR_RULES = {
         "source": "/communities/<community_id>/curate",
         "target": communities_requests_view_function,
     },
-    "redirect_communities_about": {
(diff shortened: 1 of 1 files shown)


site/zenodo_rdm/config.py
```diff
@@ -221,10 +221,6 @@
         "source": "/communities/<community_id>/curate",
         "target": communities_requests_view_function,
     },
-    "redirect_communities_about": {
```
kpsherva, line 224: not sure this was obsolete, can someone confirm, ping @alejandromumo @slint ?

slint (approved):

PR #399, opened 2023-07-03
legacy: fix Unit of Work usage for files

slint:


Pipfile.lock
```diff
@@ -897,19 +890,19 @@
...
         "invenio-access": {
             "hashes": [
-                "sha256:24385828c863f76ae7d547924b58e9a6c5da089ef24d6866dc039d9e20636483",
-                "sha256:31c9a4ce8eaaee4fea2e52701d9cc3ccc299ce718d1752c44c6832bc57e627b3"
+                "sha256:b2627255dc13dba29932e27a599a9e753f1cffa383012b581ffc213856c1aa54",
+                "sha256:d56ae0c2448b7a5c2bd85d81ae2e57ca6c65399ab3846bb9480d0911df42c565"
             ],
             "markers": "python_version >= '3.7'",
-            "version": "==1.4.5"
+            "version": "==2.0.0"
```
zzacharo, line 897: this comes with upgrading alembic see https://github.com/inveniosoftware/invenio-access/commit/2e7df149290daa5843e56a5826b7509bf3895822#diff-ea05dae3613d9af21d4d6ee71fcd6eeb5ace377181d0cf48c4b1f2cb7c7a41c6R29 and potentially migration implication...

slint: I'm thinking of skipping this one actually for now (because of the migration implication), to deploy the hotfix and update deps in a separate PR since Celery/Kombu also seems to fail...

PR #402, opened 2023-07-06
transactions stream migration

ppanero:
- Requires a release of `invenio-rdm-migrator`
- closes https://github.com/inveniosoftware/invenio-rdm-migrator/issues/109, closes https://github.com/zenodo/zenodo-rdm/issues/403


site/setup.cfg
```diff
@@ -20,12 +20,14 @@ python_requires = >=3.9
 zip_safe = False
 install_requires =
     github3.py>=3.0.0
-    invenio-rdm-migrator>=1.0.0a12,<2.0.0
     nameparser>=1.1.1
 
 [options.extras_require]
 tests =
     dictdiffer>=0.9.0
+migrator =
```
ppanero, line 28: this might be a problem since `tests/migrator` won't be able to run in CI. Shall we move them to another folder and run the manually? or in a different CI task? since after migration this module should be "removed".

site/zenodo_rdm/migrator/load.py
ppanero, line 1: Most of this loaders have been moved to rdm-migrator since they do not have any Zenodo specific code and could be used by others.

site/zenodo_rdm/migrator/models.py
ppanero, line 1: moved to rdm-migrator too

site/zenodo_rdm/legacy/deserializers/metadata.py
```diff
@@ -24,46 +24,13 @@
...
 from ..vocabularies.licenses import LEGACY_LICENSES, legacy_to_rdm
 
 record_identifiers_schemes = LocalProxy(
     lambda: current_app.config["RDM_RECORDS_IDENTIFIERS_SCHEMES"]
 )
 
-# based on
-# DOI to id https://digital-repositories.web.cern.ch/zenodo/support/operations/#openaire-grants-import
-# id to ROR: https://github.com/inveniosoftware/invenio-vocabularies/blob/master/invenio_vocabularies/config.py#L64
-FUNDER_DOI_TO_ROR = {
```
ppanero, line 36: two reasons:
1. Makes sense semantically to be in vocabularies as `LICENSES`
2. To break a dependency on invenio modules when importing it for the transform step

Pipfile
```diff
@@ -13,6 +13,7 @@ isort = "<6.0" # Unpin when isort==6.0.0 is released
 invenio-app-rdm = {version = "~=12.0.0b2.dev3", extras = ["opensearch2"]}
 invenio-logging = {extras = ["sentry-sdk"], version = "~=1.3.0"}
 zenodo_rdm = {editable="True", path="./site"}
```
ppanero, line 15: major/or nullable (if Im wrong): This will appear in the `Pipfile.lock`. Then when we run tests for migrator we doa `pipenv install`, wouldn't it cause issues with incompatible deps?

legacy/run-tests.sh
```diff
@@ -0,0 +1,23 @@
...
+
+# Quit on unbound symbols
+set -o nounset
+
+python -m check_manifest
+python -m pytest "$@"
+tests_exit_code=$?
+exit "$tests_exit_code"
+
+ls
```
ppanero, line 23: Needed?
```suggestion
```

legacy/setup.cfg
```diff
@@ -0,0 +1,36 @@
...
+universal = 1
+
+[pydocstyle]
+add_ignore = D401
+
+[isort]
+profile=black
+
+[tool:pytest]
+addopts = --black --isort --pydocstyle --cov=migrator --cov-report=term-missing
```
ppanero, line 35:
```suggestion
addopts = --black --isort --pydocstyle --cov=legacy --cov-report=term-missing
```

migrator/run-tests.sh
```diff
@@ -0,0 +1,23 @@
...
+
+# Quit on unbound symbols
+set -o nounset
+
+python -m check_manifest
+python -m pytest "$@"
+tests_exit_code=$?
+exit "$tests_exit_code"
+
+ls
```
ppanero, line 23:
```suggestion
```

migrator/setup.cfg
```diff
@@ -0,0 +1,49 @@
...
+    zenodo-legacy
+
+[options.extras_require]
+tests =
+    dictdiffer>=0.9.0
+    pytest-invenio
+    pytest-black
+
+[options.entry_points]
+# flask.commands =
```
ppanero, line 35: remove unused?
```suggestion

```

migrator/zenodo_rdm_migrator/extract/kafka.py
```diff
@@ -0,0 +1,38 @@
...
+#
+# ZenodoRDM is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+"""Migrator extraction classes for multiple sources (PGSQL, JSONL files)."""
+
+from invenio_rdm_migrator.extract import Extract
+from kafka import KafkaConsumer
+
+
+class KafkaExtract(Extract):
```
ppanero, line 13: isn't this generic enough to be in invenio_rdm_migrator?

migrator/zenodo_rdm_migrator/transform/records.py
```diff
@@ -89,3 +91,13 @@ def _transform(self, entry):
             "parent": self._parent(entry),
             "record_files": self._record_files(entry),
         }
+
+
+class ZenodoActionRecordTransform(ZenodoRecordTransform):
+    """Temporary hack until the json attribute loading is solved."""
```
ppanero, line 97: self note: for discussion, what's the issue this is solving?

ppanero (commented):
Note: cant approve since is my PR, but most comments are minor and can even be addressed in a different PR.

Note 2: I've added a few questions for discussion (for my own understanding)

Extra questions/ideas:
- Should we keep the latest "processed/seen" tx in the state? Not sure if we want to do actual state management of these (i.e. started, finished, etc.)
- Any ideas on the tx grouping? would we push back to kafka? in memory? is there another approach in mind?

legacy/setup.cfg
ppanero, line 1: The split of legacy, while keeping some things in site.zenodo_rdm.legacy  is a bit confusing (thought it makes sense)

Pipfile
slint: I've added an "if" in the GitHub action to run `pipenv install` only for the `site` test matrix, so we're covered (though a bit "hackish" and arbitrary solution 😅 )

migrator/streams.yaml
ppanero, line 1: should we add the "actions" stream?

```
action:
  extract:
    filepath:
...
```

migrator/setup.cfg
slint: Not sure if we'll eventually make this a standalone CLI and not run via `python -m zenodo_rdm_migrator` so left it in for now. Let's discuss, though it's a minor point.

migrator/zenodo_rdm_migrator/extract/kafka.py
slint: Yes... The Kerberos auth part (i.e. installing the `gssapi` Python library) is the only specific thing we need for ZenodoRDM. Though I have a feeling we will be the only ones to ever use the Kafka/Debezium pipeline for migration.

migrator/zenodo_rdm_migrator/transform/records.py
slint: This was already there, so I'm not sure either 😅

migrator/streams.yaml
slint: As a separate file though, since it's a different "migration pipeline" and we shouldn't mix the two.

migrator/zenodo_rdm_migrator/transform/records.py
ppanero: memory works: the problem is that debezium sends it as a string so we need to load it. probably a config to be changed.

PR #420, opened 2023-07-26
migrator: refactor transactions into actions

ppanero:
- closes https://github.com/inveniosoftware/invenio-rdm-migrator/issues/114

Tests passing locally with the contents of https://github.com/inveniosoftware/invenio-rdm-migrator/pull/119

Screenshot 2023-07-27 at 11 49 12


migrator/zenodo_rdm_migrator/actions/drafts.py
```diff
@@ -0,0 +1,88 @@
...
+
+from invenio_rdm_migrator.actions import TransformAction
+from invenio_rdm_migrator.load.postgresql.transactions.operations import OperationType
+from invenio_rdm_migrator.streams.actions import RDMDraftCreateAction
+from invenio_rdm_migrator.transform import IdentityTransform
+
+from ..transform.records import ZenodoRecordTransform
+
+
+class ZenodoDraftCreateAction(TransformAction):
```
ppanero, line 19: note for reviewers:

It assumes to things:
- The data returned by the extractor has already been "filtered" (e.g. removed `schema`, set `payload` content as root`)
- Some types have been changes (e.g. `op` should be an instance of `OperationType`)

migrator/tests/streams/actions/test_draft_actions_stream.py
```diff
@@ -0,0 +1,127 @@
...
+        model.__table__.create(bind=eng, checkfirst=True)
+
+    yield eng
+
+    # remove tables
+    for model in tables:
+        model.__table__.drop(eng)
+
+
+def test_draft_create_action_stream(test_extract_cls, db_engine):
```
ppanero, line 83: This sets the foundations for streams testing (and having payloads) and adds minimal checks. We could perform many more checks:
- pid status
- bucket locked or not

migrator/tests/streams/actions/test_draft_actions_stream.py
```diff
@@ -0,0 +1,127 @@
...
+
+    stream = Stream(
+        name="action",
+        extract=test_extract_cls(),
+        transform=Tx(),
+        load=PostgreSQLTx(DB_URI),
+    )
+    stream.run()
+
+    with db_engine.connect() as conn:
```
ppanero, line 95: Did not use Core because there was an issue with the declarative classes and also to distance it from the modeling used by the migrator. This way the queries are more explicit on what we check.

migrator/tests/streams/actions/test_draft_actions_stream.py
```diff
@@ -0,0 +1,127 @@
...
+                parents_state=parents_state,
+                records_state=records_state,
+                communities_state=communities_state,
+                pids_state=pids_state,
+            )
+
+    return TestExtractor
+
+
+DB_URI = "postgresql://invenio:invenio@localhost:5432/invenio"
```
ppanero, line 58: Hardcoded to docker-services-cli default values.

migrator/zenodo_rdm_migrator/actions/drafts.py
```diff
@@ -0,0 +1,69 @@
...
+            "pidstore_pid": OperationType.INSERT,
+            "files_bucket": OperationType.INSERT,
+            "records_metadata": OperationType.INSERT,
+        }
+
+        for operation in tx.operations:
+            table_name = operation["source"]["table"]
+
+            rule = rules.pop(table_name, None)
+            # rule can be None, e.g. draft creation has two rows on pidstore_pid
```
ppanero, line 38: @slint this is bad :( we need to find a better way for those "ignored ones", the use case is when there are multiple for one table and we have to ignore some of them. Could lead to errors.

ppanero:
Tests passing locally

Screenshot 2023-08-01 at 13 03 14

migrator/zenodo_rdm_migrator/stream.py
```diff
@@ -117,8 +118,8 @@
 
 ActionStreamDefinition = StreamDefinition(
     name="action",
-    extract_cls=JSONLExtract,
-    transform_cls=ZenodoTransactionGroupTransform,
-    load_cls=ZenodoTransactionLoad,
+    extract_cls=KafkaExtract,
+    transform_cls=Tx,
```
slint, line 122: minor: sorry, forgot to update this, it should point to a class extending `BaseTxTransform`, with whatever transform actions we have implanted configured in the `actions` class variable.

ppanero: Fixed, I did not export `ZenodoTxTransform` in the root module `transform` because it causes a circular import which I'm not a 100% sure how to solve (none of the options seem good to me). It's caused by the fact that the `actions/drafts.py` imports the root `ZenodoRecordTransform`.

Issue #424, opened 2023-07-27
migration-actions: User


ppanero:
- [x] Register user
    - [x] local
    - ~~[ ] GH~~ Considered as OAuth client connect action
    - ~~[ ] ORCID~~ Considered as OAuth client connect action
    - ~~AAI (Not implemented in Zenodo legacy)~~
- [x] Confirm user email: Treated as user update
- [x] Update user email
- [ ] Update user profile
- [x] Update password
- [x] New user login: Treated as user update
- [x] Misc - Automatic spam flagging (REST API)
- [x] Enable `kcat` (if not there yet). @slint  was working on this, check status.

Note: user deletion will be frozen during migration. Therefore, is not a use case to take into account.

ppanero:
From @slint (thanks!):

`kcat_qa -C -t "zenodo-migration.public" -o end`

ppanero:
Both PRs are ready to review. Note that zenodo-rdm tests are passing in local (requires the invenio-rdm-migration PR to be released).
- invenio-rdm-migrator: ⚠️ the latest commit is not in working state, it is required for the user profile partial update
- zenodo-rdm-migrator: tried to speed up tests playing with fixture scopes but still not fast enough

PR #438, opened 2023-08-02
User actions

ppanero:
- closes https://github.com/zenodo/zenodo-rdm/issues/424


migrator/tests/actions/users/test_user_actions_stream.py
ppanero, line 1: this tests interacting with the DB are becoming quite slow... but they are very much indispensable... ideas?

migrator/tests/actions/users/test_user_actions_stream.py
```diff
@@ -0,0 +1,106 @@
...
+        assert len(users) == 1
+        assert list(users)[0]._mapping["id"] == 123456
+
+        # Login information
+        loginfo = list(conn.execute(sa.select(LoginInformation)))
+        assert len(loginfo) == 1
+        assert list(loginfo)[0]._mapping["user_id"] == 123456
+
+
+def test_user_login_action_stream(
```
ppanero, line 63: for some reason user editions are not making it to the db 🤔

migrator/tests/actions/drafts/test_draft_actions_stream.py
```diff
@@ -24,25 +24,10 @@
...
-
-    return TestExtractor
-
-
 DB_URI = "postgresql+psycopg://invenio:invenio@localhost:5432/invenio"
 
 
-@pytest.fixture(scope="function")
+@pytest.fixture(scope="module")
 def db_engine():
```
slint, line 31: minor: maybe it's worth to borrow the fixture we have in `pytest-invenio` for the DB, which creates a transaction and rolls it back as clean-up? This might improve the speed issues we've discussed (and is a best-practice/recipe described in SQLAlchemy docs too)

migrator/zenodo_rdm_migrator/actions/users.py
```diff
@@ -0,0 +1,152 @@
...
+    UserDeactivationAction,
+    UserEditAction,
+    UserRegistrationAction,
+)
+from invenio_rdm_migrator.transform import IdentityTransform
+
+from ..transform.entries.users import ZenodoUserEntry
+
+
+class ZenodoUserRegistrationAction(TransformAction):
```
slint, line 23:
```suggestion
class UserRegistrationAction(TransformAction):
```

nit: to keep class names small, and maybe rely on namespacing?

migrator/zenodo_rdm_migrator/actions/users.py
```diff
@@ -0,0 +1,152 @@
...
+            if not rule or not rule == operation["op"]:
+                return False
+
+        return len(rules) == 0
+
+    def _transform_data(self):  # pragma: no cover
+        """Transforms the data and returns an instance of the mapped_cls."""
+        payload = {}
+        for operation in self.tx.operations:
+            payload = {**payload, **operation["after"]}
```
slint, line 51: minor: is this to merge all the entities into one dictionary? Aren't we afraid of overwriting/merging common keys (e.g. `updated` or `id`?).

I see that this will be a common thing, so maybe we could create some utilities on the `Tx`. class to make it easier to retrieve entries from a specific table from the `.operations` key.

ppanero: This one is done on purpose and it's meant to overwrited. The thing is this action gets multiple changes on the same row of the same table. Therefore, we only want the last one. At the moment, we could simplye do `operations[-1]` but that would be wrong if we have partial updates. With partial update the current code would end up with the sum of all changes and latest data. It's a bit "implicit". I can extract the behaviour to another function or leave a comment, however, I do not think is generic enough to move it to `Tx`. WDYT?

migrator/zenodo_rdm_migrator/actions/users.py
ppanero: Done, I've renamed classes using `Load` as prefix for RDM-Migrator ones. However, this is inconsistent with transformers since we have been using `ZenodoRecordTransform`, `ZenodoUserTransform`, etc.

migrator/tests/actions/drafts/conftest.py
```diff
@@ -0,0 +1,1976 @@
...
+# ZenodoRDM is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Migrator tests configuration."""
+
+import pytest
+from invenio_rdm_migrator.load.postgresql.transactions.operations import OperationType
+
+
+@pytest.fixture()
```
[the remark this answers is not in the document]
ppanero: They need to be function or do a deep copy since the dict gets modified. Most of the transformations are done in place, maybe not in the root dict but in some inner dicts. I think this change would improve a bit test performance, but I dont think is worth the time it requires investing in _hunting_ down all the functions that do in-place modifications. 

Extra info: I think tests were slow in my laptop for some reason (docker, or similar). In GH actions they take 20s, and now in my laptop again. When I mentioned I could not test in GH actions due to missing dependencies releases. I think we can "shelve" the performance topic.

migrator/tests/actions/drafts/test_draft_actions_stream.py
ppanero: working on this atm

ppanero: I'm quitting on this, it is very time consuming to figure it out and I dont think we will gain much in performance (it would make some test cleaner since they have atm some nasty teardown/cleanup code.

Main issues:
- It is not just a copy past from pytest-invenio
- Pytest-invenio works with flask database object and sqlalchemy 1.4
- This module works with the Engine API (not database) and SQLAlchemy 2
- I was able to make a fixture that rollsback operations executed on it, however, when we run test with `action.run()` there is a new engine and session created internally and I did not manage to make the event listener work with the engine API.

ppanero:
Tests passing locally with latest rdm-migrator

Screenshot 2023-08-09 at 13 42 29

migrator/zenodo_rdm_migrator/actions/users.py
slint: Agree... Maybe if we added some folder-level prefixing it would help, i.e. `zenodo_rdm_migrator/actions/transform/users.py`. Then we could just name the class `UserRegistration` and rely on package import aliases for referencing

migrator/zenodo_rdm_migrator/actions/users.py
slint: I don't mean moving the ops-dict merging to `Tx`, but being able to retrieve all the operations from a specific table inside a `Tx`. E.g. in this case I'm worried that if there are more tables/models modified in the operations, you end up with invalid keys for the model. 

Although I see that it's passed to `ZenodoUserEntry.transform` later on, which is only picking specific fields. Anyways, not critical now, just a bit spooky to look at :)

ppanero: I understand now thanks, I think it could be helpful. I have a similar case now with the files update action. On the other hand, I _think_ it is not possible to end up with invalid keys since if there was a different table the `match_action` would not return `True` thus we would not go into tranforming the data. But I agree this is an "implicit truth" depending on a check.

I'll open an issue, we can discuss if it is priority or not.

migrator/zenodo_rdm_migrator/actions/users.py
slint: I mean something like:

```python
from zenodo_rdm_migrator.actions import transform

actions = [
    transform.UserRegistration,
    transform.UserEdit,
    ...
]

# or
from zenodo_rdm_migrator.actions import transform as actions_transform

# kind of unnecessary though because we're defining this "actions" list, which already
# hints towards what it contains :)
actions = [  
    actions_transform.UserRegistration,
    actions_transform.UserEdit,
    ...
]
```

slint:
Once I release `invenio-rdm-migrator` I can merge this as well (I'll do a CLI merge, to avoid messing up the commit hashes).

PR #445, opened 2023-08-08
vocabularies: fix resource_types

jrcastro2:
Depends on https://github.com/inveniosoftware/invenio-rdm-records/pull/1407


app_data/vocabularies/resource_types.yaml
```diff
@@ -656,7 +656,7 @@
     openaire_type: dataset
```
jrcastro2, line 656: For this vocabulary the datacite_general attribute is Model, but we display it as an subtype of an image. To me it looks it should not be a subtype of image as if we follow the [datacite guidelines](https://support.datacite.org/docs/what-are-the-resource-types-for-datacite-dois) `model` is available, same for the openaire guidelines (https://guidelines.openaire.eu/en/latest/data/field_resourcetype.html), should it be a `dataset` or `model`?. Alternatively if we want to keep it as subtype, should we change it to:
```diff
-    datacite_general: Model
-    datacite_type: ''
+    datacite_general: Image
+    datacite_type: Model
```

slint (commented):
I think the fix here is to completely remove the file actually, since `invenio-rdm-records` has a more up to date version of it.

ppanero:
@slint @jrcastro2 I've made a diff and these are the main changes:
- zenodo-rdm has wrongly set up `ConferenceProceeding` as `publication-conferencepaper`, should be `publication-conferenceproceeding`
-zenodo-rdm has wrongly set up `model`, in rdm-records it is correct
-zenodo-rdm has wrongly set up `computationalnotebook`, in rdm-records it is correct

The problem why we cannot remove this file (unless we add some more logic to fixtures to load from more than one module, or decide is ok to have it in rdm-records), zenodo adds:

```
- id: physicalobject
  icon: cube
  props:
    csl: graphic
    datacite_general: PhysicalObject
    datacite_type: ""
    openaire_resourceType: "0010"
    openaire_type: other
    eurepo: info:eu-repo/semantics/other
    schema.org: https://schema.org/CreativeWork
    subtype: ""
    type: physicalobject
  title:
    en: Physical object
    de: Physikalisches Objekt
  tags:
    - depositable
    - linkable
```

slint (approved):

Pipfile.lock
ppanero, line 1: was the lock done inside a docker container? or on the local machine? I think you have linux in any case so it should not affect much?

jrcastro2: In a local machine

ppanero: Zach mentioned something about locking inside docker from now on, but I do not have all the details maybe Nico or Alex know more 😇

slint: @jrcastro2 can you give a go at https://github.com/zenodo/zenodo-rdm#update-dependencies and see if you get a different output? I guess there'll be no diff though since you're probably closest to the final env with Linux amd64.

jrcastro2: I did it inside the container and got the almost the same output, only change was this one:
```console
javier@javier-laptop:~/cern/rdm-demo-sites/zenodo-rdm$ git diff
diff --git a/Pipfile.lock b/Pipfile.lock
index 80b9e94..42c5406 100644
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -2746,7 +2746,7 @@
                 "sha256:f9062af71c59c004cd519e2fb8f5d25d39e46d3af011b41ab43b9c74e27e236f",
                 "sha256:fb074d8b213749fa1d077d630db0d5f8cc3b2ae63587ad4116e8a436e9bbe995"
             ],
-            "markers": "python_version >= '3.6'",
+            "markers": "python_version >= '3.8'",
             "version": "==23.7.0"
         },
         "blinker": {
```

Issue #429, opened 2023-08-01
migration-actions: Files


slint:
- [ ] Create/upload a file (old) (REST API)
- [ ] Delete a file (old) (REST API)
- [ ] Replace a file (old) (REST API)
- [ ] Create/upload a file (new) (REST API)
- [ ] Delete a file (new) (REST API)
- [ ] Replace a file (new) (REST API)
- [ ] File-integrity check (Celery)

PR #446, opened 2023-08-09
migration: file upload action

ppanero:
- requires https://github.com/zenodo/zenodo-rdm/pull/438
- requires https://github.com/inveniosoftware/invenio-rdm-migrator/pull/129
- partly closes https://github.com/zenodo/zenodo-rdm/issues/429


ppanero:
Tests passing locally with latest changes and latest invenio-rdm-migrator (not released yet)

Screenshot 2023-08-11 at 16 16 29

migrator/zenodo_rdm_migrator/actions/transform/files.py
```diff
@@ -0,0 +1,85 @@
...
+        """Checks if the data corresponds with that required by the action.
+
+        Should contain two updates on the bucket (updated, size), the creation
+        and the link to the file_id (update) on the object version, and the creation and commit
+        (update) of the file instance.
+        """
+        rules = {
+            "files_bucket": [OperationType.UPDATE, OperationType.UPDATE],
+            "files_object": [OperationType.UPDATE, OperationType.INSERT],  # lifo
+            "files_files": [OperationType.UPDATE, OperationType.INSERT],  # lifo
```
slint, line 33: minor/shelve: there is an "edge" case for file upload via REST API, where you can "overwrite" a file with the same name. This effectively is an update to the old ObjectVersion (to set `is_head = False`) and then a creation of a new ObjectVersion + FileInstance.

migrator/zenodo_rdm_migrator/actions/transform/files.py
```diff
@@ -0,0 +1,85 @@
...
+                fi = {**fi, **operation["after"]}
+
+        fr = {
+            "id": None,  # generated by the load action
+            "json": {},
+            "created": ov["created"],
+            "updated": ov["updated"],
+            "version_id": 1,
+            "key": ov["key"],
+            "record_id": None,  # calculated by the load action
```
slint, line 76: minor/comment: do we have a rule of thumb over who is responsible for the "linking" to existing objects? IIRC, we do have access to the state in the Transform In this case, so we could fish the record ID using the bucket ID.

slint (approved):
LGTM! Two minor comments, one can be shelved as a new list item in the original issue, since it's an edge case.

migrator/zenodo_rdm_migrator/actions/transform/files.py
ppanero: I will add a comment to the epic issue, I think in general we will have to revise the way we do the fingerprinting for edge cases

migrator/zenodo_rdm_migrator/actions/transform/files.py
ppanero: I was also divided at heart on this one. I decided to keep it in the load, since it is how we do for other cases in drafts. 

For example, both in snapshot and action migration we might have to check if it is a draft of a published version or not; then we have to use one data or another. At the moment this is taken from the state and is done in load.

In my opinion, it is the same type of operation: _calculate a value based on the state_. It falls again into is it business logic/data transformation of is it data processing? To be consistent I kept it in the `load`, but I'm not against moving all this operations to transform. I think it makes sense:
- load: get data and db interaction
- tranform: do all transformation (from data and state)

⚠️ this would be a big refactor 😢

slint: Thanks for the mental walkthrough :)

I retraced a bit what we were discussing at the very beginning of the ETL pipeline and I think it's good practice to delegate to the Load. 

The logic is that you can feed to the Load step data that has "holes" (e.g. missing PKs, relations/FKs) and it will take the responsibility of either "minting" new PKs or figuring out if/what is the related object FK. With that in mind, in this example the data we're passing to Load is saying "there was a file uploaded to bucket X, that is connected to a record; please figure out which record that is".

What I would maybe make a bit more clear throughout the classes is that this was a `DraftFileUpload` action (and not just any `FileUpload` action). I think with the type of actions it's generally better to be very specific (since they have to represent real-life possible actions on either legacy or RDM). This doesn't mean of course that the underlying code cannot be generic and shared (i.e. no need to duplicate code for the Transform/Load actions).

PR #450, opened 2023-08-10
migrator: use microseconds for datetimes instead of str

ppanero:
- undo https://github.com/zenodo/zenodo-rdm/commit/637dc4af62857e622b3a2d262795a853051bec86
- requires https://github.com/inveniosoftware/invenio-rdm-migrator/pull/131

Tests passing locally

Screenshot 2023-08-10 at 16 35 14

### DIFF
--- a/migrator/tests/actions/drafts/test_draft_actions_stream.py
+++ b/migrator/tests/actions/drafts/test_draft_actions_stream.py
@@ -58,14 +58,20 @@ def test_draft_create_action_stream(
         drafts = list(conn.execute(sa.select(RDMDraftMetadata)))
         assert len(drafts) == 1
 
+        draft = drafts[0]._mapping
+        assert draft["created"] == "2023-06-29T13:38:48.842023"
+        assert draft["updated"] == "2023-06-29T13:38:48.842023"
+
         parents = list(conn.execute(sa.select(RDMParentMetadata)))
         assert len(parents) == 1
+
+        parent = parents[0]._mapping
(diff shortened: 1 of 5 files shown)


migrator/zenodo_rdm_migrator/transform/entries/records/records.py
```diff
@@ -26,11 +26,17 @@ def _id(self, entry):
 
     def _created(self, entry):
         """Returns the creation date of the record."""
-        return entry["created"]
+        value = entry["created"]
```
ppanero, line 29: It was very complex to create subclasses of entries and overwrite `Transform` classes just to modify date values. Since there is a clear distinction between snapshot (comes as string) and actions (comes as microseconds/int), this implementation should suffice.

migrator/tests/actions/drafts/test_draft_actions_stream.py
ppanero, line 1: added dates checking to the tests since the DB is type string and was allowing everything to pass (e.g. integers being casted, etc.). We cannot be less permissive (e.g. change type to datetime) because the models are also used for snapshot migration and they come as string.

migrator/zenodo_rdm_migrator/actions/users.py
ppanero, line 1: This code becomes a bit cumbersome, I'm not convinced. However, a bit part of it will dissapear/change once the updates (e.g. edit) become partial.

slint (approved):

PR #454, opened 2023-08-14
migrator: transactions Kafka extract

### LINKED ISSUE #447: migration: implement dict-diffing in Kafka Extract for UPDATE
Since we now have both the `before` and `after` state of a statement, we can filter out the unchanged fields in the payloads to make it easier for the transformer to apply the data changes.

slint:
* Closes https://github.com/inveniosoftware/invenio-rdm-migrator/issues/113.
* Closes #447.


migrator/tests/extract/test_kafka_extract.py
```diff
@@ -0,0 +1,260 @@
...
+    tx_op_counts=None,
+):
+    assert len(result) == count
+    assert all(isinstance(t, Tx) for t in result)
+    tx_dict = {t.id: t for t in result}
+    tx_ids = list(tx_dict.keys())
+    assert tx_ids[0] == first_tx_id
+    assert tx_ids[-1] == last_tx_id
+    assert set(extra_tx_ids or []) <= set(tx_ids)
+    assert len(set(excluded_tx_ids or []).intersection(set(tx_ids))) == 0
```
ppanero, line 110: sets do not guarantee order, does this always pass?

migrator/zenodo_rdm_migrator/extract/kafka.py
```diff
@@ -7,33 +7,291 @@
...
+        """Get transaction info."""
+        return self._info
+
+    @info.setter
+    def info(self, val):
+        """Set transaction info."""
+        self._info = val  # store the raw value for debugging
+        if val:
+            # We only care about the table row counts stated in the transaction event
+            self._info_counts = Counter(
```
ppanero, line 43: 🚀

migrator/zenodo_rdm_migrator/extract/kafka.py
```diff
@@ -7,33 +7,291 @@
...
+from datetime import datetime
+from pathlib import Path
+
+from invenio_rdm_migrator.extract import Extract, Tx
+from invenio_rdm_migrator.logging import Logger
+from kafka import KafkaConsumer, TopicPartition
+from sortedcontainers import SortedDict, SortedList
+
+
+class TxState:
```
ppanero, line 21: Should this be moved along with `Tx`? or otherwise be called `KafkaTxState`, if it is very tied to kafka behaviour?

migrator/zenodo_rdm_migrator/extract/kafka.py
```diff
@@ -7,33 +7,291 @@
...
+            yield Tx(id=tx.id, operations=list(tx.ops))
 
     def run(self):
-        """Yield one element at a time."""
-        consumer = KafkaConsumer(*self.topics, **self.kafka_cfg)
-        yield from consumer
+        """Return a blocking generator yielding completed transactions."""
+        # We're using an (always) SortedDict, since we want to yield transactions in
+        # their correct order.
+        try:
```
ppanero, line 275: understanding check:
1. we consume all available info (does this ever stop?)
2. we consume all the available ops
3. we yield those that completed
4. back to step 1

I wonder if there is risk of starvation in any of the steps (i.e. not passing to the subsequent).

migrator/zenodo_rdm_migrator/extract/kafka.py
```diff
@@ -7,33 +7,291 @@
...
+                    tx_state = self.tx_registry.setdefault(tx_id, TxState(tx_id))
+                    tx_state.append(op)
+                    if tx_state.complete:
+                        self.logger.info(f"Completed transaction {tx_state.id}")
+
+                # If no new transactions, we don't need to sleep since consumers
+                # have a timeout/sleep already via "consumer_timeout_ms".
+                yield from self._yield_completed_tx(min_batch=self.tx_buffer)
+        # Normally this extract would run forever, so we need some mechanism to stop
+        except KafkaExtractEnd:
```
ppanero, line 295: understanding check: when would we hit this? at some point we would "switch off" old Zenodo and then this would happen. That's it right?

ppanero (approved):
At first pass it LGTM! tbh, it is not trivial, I'd need to spend lots of time to understand what's going on at every step so I just made a few comments to make sure I understood and that there's no starvation.

Maybe it's worth leaving a comment on why we need two consumers (tx, ops), I mean on how the topics/partitions are expected (so that info is not only in the ServiceNow ticket).

Feel free to merge!

migrator/tests/extract/test_kafka_extract.py
slint: In this case, it's only testing exclusion of certain transactions, and not the order. 

Overall order is guaranteed though, because of the core logic of how the `KafkaExtract` yields transactions: the `self.tx_registry` is a `sortedcontainers.SortedDict`, which ensures insertion order of keys and yielded items are "popped" from the beginning. 

In any case, I should add a "sanity check" though of the results order :)

migrator/zenodo_rdm_migrator/extract/kafka.py
slint: It's a bit simplistic, but that's the only info we have also from Debezium :)

migrator/zenodo_rdm_migrator/extract/kafka.py
slint: Yeah, it's more like a "pending transaction container" that helps build it and easily check if a transaction is "complete". Originally I called it `PendingTx` (but felt wrong to check for `pending_tx.complete`)... It's only used internally, so either prefixing with `KafkaTxState` or making it private (e.g. `_TxState`) would also work.

migrator/zenodo_rdm_migrator/extract/kafka.py
slint: > 1. we consume all available info (does this ever stop?)

Correct. Once we go through the "bulk" of it (depends on the starting point/offset/timestamp), it starts behaving more like a live stream that returns continuous batches (it stops iterating if there are no messages after 3sec, as per the `consumer_timeout_ms` config). 

I'm actually considering adding a "max messages" param to chunk the delta runs, because the first time we run a delta (i.e. after the snapshot has been loaded) it will be ~8h worth of messages, which will be stored in memory... So pretty good question/catch :)

> 2. we consume all the available ops
> 3. we yield those that completed

important: we yield the "earliest" transactions that have been completed. I.e. imagine:

1. We got Tx1, Tx2, and Tx3
2. Tx2 and Tx3 are complete, but Tx1 still has ops (that will come in a later iteration of the consumer)
3. In this case we don't yield Tx2 and Tx3, since we're still waiting for Tx1 to complete (hopefully in the next iteration).

> 4. back to step 1

Correct!

> I wonder if there is risk of starvation in any of the steps (i.e. not passing to the subsequent).

See the point regarding the `consumer_timeout_ms` above ☝️

migrator/zenodo_rdm_migrator/extract/kafka.py
slint: Exactly, when we shut down legacy Zenodo, we'll just stop seeing new transactions coming in. At that point we hit Ctr-C (or some other mechanism that throws `KafkaExtractEnd`, e.g. creating a file in a directory, or some other process signaling). 

For now, this is only used in tests, otherwise there's no way to "stop" the `KafkaExtract.run()` iterator.

migrator/zenodo_rdm_migrator/extract/kafka.py
ppanero: This is very helpful, thanks for taking the time to write details comments. I wonder if we should put this in the class docstring? since it is not trivial to understand it might help others.

slint: Added the max fetching parameters, and some additional docstring comments in https://github.com/zenodo/zenodo-rdm/pull/454/commits/d93da36210d4211efeb7146752a965bf60eda62a

migrator/zenodo_rdm_migrator/extract/kafka.py
slint: Renamed in the latest commit to `_TxState`.

slint:
I pushed the dict-diffing here as well with a test in https://github.com/zenodo/zenodo-rdm/pull/454/commits/3702edc5f36b9fdad56ce103ced2277d3adc7ace

migrator/zenodo_rdm_migrator/extract/kafka.py
```diff
@@ -7,33 +7,348 @@
...
+            elif tx_msg.value["status"] == "END":
+                consumer.commit()
+                yield (tx_id, tx_msg.value)
+
+    @staticmethod
+    def _filter_unchanged_values(msg):
+        # We don't have deeply nested dict values, since they are still raw JSON strings
+        before = msg.value["before"]
+        after = msg.value["after"]
+        pk_keys = set(msg.key.keys())
```
ppanero, line 254: self note: I assume this `.key` with the pks is something that kafka/debezium gives? I see it's being extracted from `op_msg.key` coming from the consumer messages.

ppanero (approved):

PR #465, opened 2023-08-24
profiler: initial module

slint:
* Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2361

https://github.com/inveniosoftware/invenio-app-rdm/assets/1814661/85982f58-f4d8-4f95-a8d6-997773eb48e9

### DIFF
--- a/site/zenodo_rdm/templates/semantic-ui/profiler/index.html
+++ b/site/zenodo_rdm/templates/semantic-ui/profiler/index.html
@@ -1,0 +1,115 @@
+{%- extends config.BASE_TEMPLATE %}
+
+{% block title %}Profiler{% endblock title %}
+
+{% block page_body %}
+<div class="ui container">
+  <div class="ui grid">
+    <div class="two column row">
+      <div class="five wide column">
+        <h2>Profiler</h2>
+        {% if active_session %}
+        <dl>
+          <dt class="ui header">Session ID</dt>
+          <dd>{{ active_session.id }}</dd>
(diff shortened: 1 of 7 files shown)


slint:
I have no good quick solutions on where to place this code... Some options:

- New module `Flask-Profiling` (but takes some time)
- Part of `invenio-app(-rdm)` with extras... Usage though will be very "manual" (i.e. you install the extras yourself, and initialize the extension yourself as well)

site/zenodo_rdm/templates/semantic-ui/profiler/index.html
```diff
@@ -0,0 +1,115 @@
...
+    </div>
+  </div>
+</div>
+
+{% endblock page_body %}
+
+{% block javascript %}
+{{ super() }}
+<script type="text/javascript">
+  $('.ui.accordion').accordion();
```
kpsherva, line 112: it was not working without it ? we might be missing js assets for semantic ui then... (not the react ones)

slint: Yeah, was also surprised that I had to call it manually... I just noticed we have [this guy](https://github.com/inveniosoftware/invenio-theme/blob/d1f59e10634f5f9ddb5450bed71b6aa685ec0381/invenio_theme/assets/semantic-ui/js/invenio_theme/theme.js#L45-L61), which might be changing a bit the behavior (i.e. what component toggles the accordion).

PR #476, opened 2023-09-04
Migration run changes

slint:
- docs: formatting
- migrator: adjust for new "owned_by" format
- migrator: fix communities handling for drafts
- migrator: add user safelisting
- migrator: swtich to .sql extract scripts
- migrator: clean community descriptions
- migrator: add community OAI sets
- global: handle legacy recrods with no parent DOI
- migrator: fix constraint generation scripts
- migrator: implement GitHub stream
- migrator: add missing indices script
- migrator: import fixes
- migrator: parents fix
- migrator: fix draft transform
- migrator: dump media bucket IDs from "extra_formats"
- migrator: fix drop indices script

### DIFF
--- a/migrator/tests/actions/conftest.py
+++ b/migrator/tests/actions/conftest.py
@@ -55,7 +55,7 @@ def state(tmp_dir):
     state_db = StateDB(
         db_dir=tmp_dir.name, validators={"parents": ParentModelValidator}
     )
-    STATE.initialized_state(state_db)
+    STATE.initialized_state(state_db, cache=False)
 
     return STATE
 
--- a/migrator/tests/actions/drafts/test_drafts_actions_stream.py
+++ b/migrator/tests/actions/drafts/test_drafts_actions_stream.py
@@ -58,6 +58,7 @@ def db_draft(database, session):
             version_id=1,
             index=1,
             bucket_id="0e12b4b6-9cc7-46df-9a04-c11c478de211",
+            media_bucket_id=None,
             parent_id="9493793c-47d2-48e2-9867-ca597b4ebb41",
             expires_at="2024-05-01T00:00:00",
             fork_version_id=None,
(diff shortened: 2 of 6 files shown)


migrator/zenodo_rdm_migrator/transform/entries/records/records.py
```diff
@@ -14,7 +14,7 @@
 from .custom_fields import ZenodoCustomFieldsEntry
 from .metadata import ZenodoDraftMetadataEntry, ZenodoRecordMetadataEntry
 
-ZENODO_DATACITE_PREFIX = "10.5281"
+ZENODO_DATACITE_PREFIX = "10.5281/"
```
ppanero, line 17: where is this used? https://github.com/search?q=repo%3Azenodo%2Fzenodo-rdm%20DATACITE_PREFIX&type=code

migrator/zenodo_rdm_migrator/transform/entries/users.py
```diff
@@ -47,6 +47,16 @@ def _confirmed_at(self, entry):
         """Returns the confirmation date."""
         return entry.get("confirmed_at")
 
+    def _blocked_at(self, entry):
+        """Returns the blocking/inactivation date."""
+        if not entry["active"]:
+            return datetime.utcnow().isoformat()
```
ppanero, line 53: This might make us loose some insight on when the blocking happened. What about using the last login information? (and if not available, default to now)

ppanero (approved):

migrator/zenodo_rdm_migrator/transform/entries/users.py
slint: Agree, didn't think of it at the time :)

Yeah, I can use this date or worst case even the email confirmation date.

PR #496, opened 2023-09-12
requests: implement a new community manage record request

anikachurilova:
* closes https://github.com/zenodo/rdm-project/issues/243

Request details page:
Screenshot 2023-09-12 at 16 17 37

Accepted:
Screenshot 2023-09-12 at 16 18 39

Declined:
Screenshot 2023-09-12 at 16 19 10

Expired:
Screenshot 2023-09-12 at 17 01 50

### DIFF
--- a/site/zenodo_rdm/legacy/requests/community_manage_record.py
+++ b/site/zenodo_rdm/legacy/requests/community_manage_record.py
@@ -143,7 +143,7 @@ class DeclineAction(actions.DeclineAction):
                 legacy_record["parent"].get("communities", None).get("ids", None)
             )
             if communities is not None:
-                not_my_communities = []
+                my_communities = []
                 for community_id in communities:
                     com_members = current_communities.service.members.search(
                         system_identity,
@@ -151,13 +151,15 @@ class DeclineAction(actions.DeclineAction):
                         extra_filter=dsl.query.Bool(
                             "must",
                             must=[
-                                dsl.Q("term", **{"role": "owner"}),
-                                ~dsl.Q("term", **{"user_id": current_user.id}),
+                                dsl.Q("term", **{"role": "owner"})
+                                | dsl.Q("term", **{"role": "curator"}),
+                                dsl.Q("term", **{"user_id": current_user.id}),
                             ],
                         ),
                     )
                     if com_members.total > 0:
-                        not_my_communities.append(community_id)
+                        my_communities.append(community_id)
+                not_my_communities = list(set(communities) - set(my_communities))
                 _remove_record_from_communities(record, not_my_communities)
             _remove_permission_flag(record, uow)
 
--- a/site/zenodo_rdm/legacy/requests/community_manage_record.py
+++ b/site/zenodo_rdm/legacy/requests/community_manage_record.py
@@ -1,0 +1,211 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 CERN.
+#
+# Zenodo is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Zenodo community manage record request."""
+
+from flask_login import current_user
+from invenio_access.permissions import system_identity
+from invenio_communities.proxies import current_communities
+from invenio_pidstore.models import PersistentIdentifier
+from invenio_rdm_records.proxies import (
+    current_rdm_records_service,
+    current_record_communities_service,
+)
+from invenio_rdm_records.records import RDMParent, RDMRecord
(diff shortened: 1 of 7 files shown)


site/zenodo_rdm/legacy/requests/utils.py
```diff
@@ -82,3 +83,26 @@ def submit_record_upgrade_request(record, uow, comment=None):
...
         system_identity, request_item._request.id, "submit", data=comment, uow=uow
     )
+
+
+@unit_of_work()
+def submit_community_manage_record_request(user_id, uow, comment=None):
+    """Create and submit a CommunityManageRecord request."""
+    type_ = current_request_type_registry.lookup(CommunityManageRecord.type_id)
+    receiver = ResolverRegistry.resolve_entity_proxy({"user": user_id}).resolve()
+    expires_at = datetime.utcnow() + timedelta(weeks=20)
```
anikachurilova, line 93: Lars' comment about the expiration date not being during Christmas holidays: just added 5 months, so that it will just be the end of January

site/zenodo_rdm/legacy/requests/utils.py
```diff
@@ -82,3 +83,26 @@ def submit_record_upgrade_request(record, uow, comment=None):
     return current_requests_service.execute_action(
         system_identity, request_item._request.id, "submit", data=comment, uow=uow
     )
+
+
+@unit_of_work()
+def submit_community_manage_record_request(user_id, uow, comment=None):
```
anikachurilova, line 89: Request is created per each `user_id`. 
Question: the request should not be created for those users who don't have any records yet. Should this chech happen here or it will be applied only for needed users somehow during migration? ping @slint

site/zenodo_rdm/legacy/requests/community_manage_record.py
```diff
@@ -0,0 +1,211 @@
...
+            if communities is not None:
+                not_my_comm = []
+                for community_id in communities:
+                    com_members = current_communities.service.members.search(
+                        system_identity,
+                        community_id,
+                        extra_filter=dsl.query.Bool(
+                            "must",
+                            must=[
+                                dsl.Q("term", **{"role": "owner"}),
```
anikachurilova, line 154: Do we know if it has to be only the owner or curator as well? ping @slint

site/zenodo_rdm/legacy/requests/community_manage_record.py
```diff
@@ -0,0 +1,211 @@
...
+                not_my_communities = []
+                for community_id in communities:
+                    com_members = current_communities.service.members.search(
+                        system_identity,
+                        community_id,
+                        extra_filter=dsl.query.Bool(
+                            "must",
+                            must=[
+                                dsl.Q("term", **{"role": "owner"}),
+                                ~dsl.Q("term", **{"user_id": current_user.id}),
```
anikachurilova, line 155: Got one more question during testing: in case a community has multiple owners: me and not me, is the record still removed from it?

site/zenodo_rdm/legacy/requests/community_manage_record.py
slint: Curator has also `edit` rights, so I think it should be both... Once we migrate there won't be multiple members yet in communities, but eventually there might, so it's good to take into account.

site/zenodo_rdm/legacy/requests/community_manage_record.py
slint: If the record owner is also an owner of the community, the record should not be removed (though see my point above also about multiple members)

site/zenodo_rdm/legacy/requests/utils.py
slint: Users without records don't matter in this case, since there won't be any permission flags added to query against (i.e. the query will return no results).

slint (approved):
LGTM! Just calrifications, no changes needed.

site/zenodo_rdm/legacy/requests/utils.py
anikachurilova: Yes that's true, but my point is, that the request will still submitted for those users (i didn't add any checks here because was not sure if it's done in migration, but if you plan to just apply this for all the users, i should add something)

slint: Ah ok, I understood now. 

So don't worry, when we run the script that calls this function to create the requests, we'll take that into account there, so no need to have extra checks in this code (just to keep it simpler).

anikachurilova: Ok, thank you for clarifying!

PR #500, opened 2023-09-13
metrics: Added the metrics module from legacy

yashlamba:
Moved: https://github.com/zenodo/zenodo/tree/master/zenodo/modules/metrics from Zenodo -> ZenodoRDM

Quite a few indices for opensearch and fields have been updated.

Closes https://github.com/zenodo/rdm-project/issues/60


site/zenodo_rdm/metrics/api.py
```diff
@@ -0,0 +1,130 @@
...
+                index=build_alias_name("stats-file-download"),
+            )
+            .filter(
+                "range",
+                timestamp=time_range,
+            )
+            .filter(
+                "term",
+                is_parent=False,
+            )
```
yashlamba, line 45: Hey @slint, need some context for this query. I don't think we have this field in ES anymore, is there a different field I should consider? Maybe alternate query is `parent_recid` should exist?

slint: I might be that actually we're not calculating stats the same way for all version as we did in legacy. Which means that you might not need this filter anymore. 

The difference between legacy/RDM stats aggregations for a record with 3 versions, is that we were also indexing a document containing aggregated numbers for all versions of a record (which is not the case anymore).

I would test to be sure that if e.g. you have 3 records with 2 file downloads each, we don't "double-count" everything.

slint: > @slint I checked, if we do download all, then only we count all files otherwise each file download is individually added.

Ok, that sounds good. The download-all is a new feature in RDM, so it will only apply there once we start getting usage.

site/zenodo_rdm/metrics/utils.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def formatted_response(metrics):
+    """Format metrics into Prometheus format."""
+    response = ""
+    for metric in metrics:
+        response += (
+            "# HELP {name} {help}\n# TYPE {name} {type}\n{name} "
+            "{value}\n".format(**metric)
+        )
```
slint, line 45:
```suggestion
        response += (
            "# HELP {name} {help}\n# TYPE {name} {type}\n{name} "
            "{value}\n"
        ).format(**metric)
```

nit: just to be sure we don't accidentally format only part of the string.

site/zenodo_rdm/metrics/api.py
```diff
@@ -0,0 +1,126 @@
...
+                timestamp=time_range,
+            )
+            .params(request_timeout=120)
+        )
+        search.aggs.metric("download_volume", "sum", field="volume")
+        result = search[:0].execute().aggregations.to_dict()
+        download_volume = result.get("download_volume", {}).get("value", 0)
+
+        search = (
+            Search(using=current_search_client, index=build_alias_name("rdmrecords"))
```
slint, line 49:
```suggestion
            Search(using=current_search_client, index=build_alias_name("rdmrecords-records"))
```

To ensure we query only published records (I think the top-level `rdmrecords` alias also includes draft indices).

site/zenodo_rdm/metrics/ext.py
```diff
@@ -0,0 +1,38 @@
...
+    def __init__(self, app=None):
+        """Extension initialization."""
+        if app:
+            self.init_app(app)
+
+    @staticmethod
+    def init_config(app):
+        """Initialize configuration."""
+        for k in dir(config):
+            if k.startswith("METRICS"):
```
slint, line 27:
```suggestion
            if k.startswith("METRICS_"):
```

A mistake from the original module as well I guess ;)

slint (approved):
Some minor comments, but otherwise looks good!

Regarding commit message style, try to keep a single component in the message, also for brevity, i.e.

```diff
-site: zenodo_rdm: metrics: imported metrics module from zenodo/zenodo
+metrics: imported metrics module from zenodo/zenodo
```

Sometimes if we have "sub-components" we use dashes, e.g. `deposit-ui: fixed validation error display`.

yashlamba:
Hey @slint, thanks for the review. Before we merge this, we should add + test the Uptime bot secrets required as well.

PR #522, opened 2023-09-19
stats: move Piwik stats exporter (stats module) from legacy

yashlamba:
Closes https://github.com/zenodo/rdm-project/issues/69

Moved the PiwikExporter from the stats module from legacy Zenodo to RDM. Have tested export to Piwik.

There's a cache field `piwik_export:bookmark` that we need to setup as the code doesn't (and I guess shouldn't) handle the None case for.


site/zenodo_rdm/stats/exporters.py
```diff
@@ -0,0 +1,146 @@
...
+from zenodo_rdm.stats.utils import chunkify, fetch_record
+
+
+class PiwikExporter:
+    """Events exporter."""
+
+    def run(self, start_date=None, end_date=None, update_bookmark=True):
+        """Run export job."""
+        if start_date is None:
+            bookmark = current_cache.get("piwik_export:bookmark")
```
slint, line 32: note/shelve: we need to remember to "migrate" this bookmark value from the legacy system to production (so that we don't resend statistics from the very beginning).

But we'll also be reminded by the logged warning below :)

site/zenodo_rdm/stats/utils.py
```diff
@@ -0,0 +1,37 @@
...
+"""Statistics utilities."""
+
+import itertools
+
+from invenio_rdm_records.proxies import current_rdm_records_service
+
+try:
+    from functools import lru_cache
+except ImportError:
+    from functools32 import lru_cache
```
slint, line 17:
```suggestion
from functools import lru_cache
```

I think this was a Python 2/3 compatibility trick, so not needed anymore

slint (approved):
LGTM! Couple of comments for removing dead code, but otherwise it's good to merge ✨

site/zenodo_rdm/stats/config.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 CERN.
+#
+# ZenodoRDM is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Configuration for ZenodoRDM stats."""
+
+STATS_PIWIK_EXPORTER = {
```
yashlamba, line 10: Question: @slint Should this be in `invenio.cfg`? I don't think it matters since env vars will take precedence.

slint: It's ok for it to stay here, also to keep the module/logic somewhat isolated. Indeed we'll override it eventually via environment on OpenShift.

site/zenodo_rdm/stats/config.py
```diff
@@ -0,0 +1,21 @@
...
+    "id_site": 1,
+    "url": "https://analytics.openaire.eu/piwik.php",
+    "token_auth": "api-token",
+    "chunk_size": 50,  # [max piwik payload size = 64k] / [max querystring size = 750]
+}
+
+STATS_PIWIK_EXPORT_ENABLED = True
```
slint, line 17:
```suggestion
STATS_PIWIK_EXPORT_ENABLED = False
```

Let's disable by default, so that e.g. local instances don't run the task and unsuccessfully try to send statistics to Piwik.

site/zenodo_rdm/stats/config.py
```diff
@@ -0,0 +1,21 @@
...
+
+# Queries performed when processing aggregations might take more time than
+# usual. This is fine though, since this is happening during Celery tasks.
+STATS_ELASTICSEARCH_CLIENT_CONFIG = {"timeout": 60}
```
slint, line 21:
```suggestion
```

I think this is also a remnant config that is not used by the stats exporter, so we can remove it.

slint (approved):
LGTM! Small changes to improve local development, and we can merge

PR #535, opened 2023-09-22
global: enabled sort by verified

alejandromumo:


site/zenodo_rdm/redirector.py
```diff
@@ -54,7 +54,9 @@ def communities_detail_view_function():
     :return: url for the view 'invenio_app_rdm_communities.communities_detail'
     :rtype: str
     """
-    _id = request.view_args.get("id", request.view_args["community_id"])
+    _id = request.view_args.get("id")
```
alejandromumo, line 57: the default value being analysed when it wasn't needed.  This was throwing an exception when `community_id` was not in `view_args`

invenio.cfg
```diff
@@ -550,3 +550,12 @@ OPENAIRE_DIRECT_INDEXING_ENABLED = True
 """Enable sending published records for direct indexing at OpenAIRE."""
 
 RDM_RECORDS_SERVICE_COMPONENTS = DefaultRecordsComponents + [OpenAIREComponent]
+
+RDM_SEARCH_SORT_BY_VERIFIED = True
+"""Enable the sorting of records by verified"""
```
kpsherva, line 555: missing a dot, not sure why black formatter didn't complain ?

site/zenodo_rdm/redirector.py
kpsherva: do we know in which cases we end up with missing community_id while the code expected it? Does it make sense to identify why it happens?

alejandromumo: Indeed, by having a closer look  I found out that `community_id` was never a view argument for any of the redirected routes (the pydoc was misleading). 

The redirected routes that hit this view are the following:

```python
    "redirect_communities_about_legacy": {
        "source": "/communities/about/<id>/",
        "target": communities_detail_view_function,
    },
    "redirect_collections_about": {
        "source": "/collection/user-<id>",
        "target": communities_detail_view_function,
    },
```

which means that `id` is the only view argument.

kpsherva (approved):

PR #537, opened 2023-09-22
metrics: evaluation & result improvements

yashlamba:
This PR aims to complete and improve the metrics module

- [x] Implement cache/flag to avoid request/task pilling for metrics calculation when cache expires
- [ ] Send some (TBD) values when an error occurs while evaluating a metric (try to be prometheus friendly)

### DIFF
--- a/site/zenodo_rdm/metrics/views.py
+++ b/site/zenodo_rdm/metrics/views.py
@@ -9,6 +9,7 @@
 
 import humanize
 from flask import Blueprint, Response, current_app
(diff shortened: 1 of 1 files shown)


slint (approved):
LGTM! Isort is complaining on tests, but otherwise once fixed we can merge 💪

slint:
Merged for now. Regarding

> Send some (TBD) values when an error occurs while evaluating a metric (try to be prometheus friendly)

Let's update once we get an answer from Marek.

Issue #524, opened 2023-09-20
infra: update configuration for the different environments


slint:
We currently have a bit of a mix/conflict between configuration sources:

- this repo's `invenio.cfg`
- GitLab repository with OpenShift Helm charts values: https://gitlab.cern.ch/zenodo/zenodo-rdm-openshift/
- What is actually the deployed/live config on the various OpenShift projects (ConfigMaps and Secrets)

Given that multiple new features require specialized secrets/config (https://github.com/zenodo/zenodo-rdm/pull/500, https://github.com/zenodo/zenodo-rdm/pull/522, OpenAIRE direct indexing), we should take the chance to reconcile all these config sources.

Some important notes:

- Application config (available even in local development) should be only in `invenio.cfg` (or even in upstream modules if it makes sense IvnenioRDM-wide). One such example is the `GITHUB_WEBHOOK_RECEIVER_ID = 'gtihub'` (we would never use a different value in dev/prod/QA).
- Ideally, we should always aim at using `helm upgrade` (or similar) to apply any configuration changes to the deployed environments
- Synchronize with @psaiz as well for changes to the GitLab repo, since there is some ongoing parallel work on OpenShift infra at the moment

alejandromumo:
I've found that the following configs could be removed from openshift:

- `GITHUB_WEBHOOK_RECEIVER_ID`  (removed from zenodo-rdm-openshift, it's added by the module and should work)
- `GITHUB_INTEGRATION_ENABLED` (removed from zenodo-rdm-openshift, already exists in zenodo_rdm/invenio.cfg)
- `DATACITE_ENABLED` (removed from zenodo-rdm-openshift, already exists in zenodo_rdm/invenio.cfg)
- `DATACITE_PREFIX` (removed from zenodo-rdm-openshift, already exists in zenodo_rdm/invenio.cfg)
- `GITHUB_RELEASE_CLASS` (removed from openshift, already in invenio.cfg and added a default in invenio-app-rdm)
- `MAIL_DEFAULT_SENDER` (removed from openshift, moved to invenio.cfg)
- `ZENODO_RECORDS_UI_CITATIONS_ENDPOINT` (removed from openshift, already in zenodo_rdm/config.py)

PR #542, opened 2023-09-25
configs: cleanup invenio.cfg

alejandromumo:
closes https://github.com/zenodo/zenodo-rdm/issues/524

### DIFF
--- a/invenio.cfg
+++ b/invenio.cfg
@@ -85,11 +85,14 @@ APP_DEFAULT_SECURE_HEADERS = {
             "'unsafe-inline'",  # for inline scripts and styles
             "'unsafe-eval'", # for webpack build
             "blob:",            # for pdf preview
+            "zenodo-broker.web.cern.ch",
             "zenodo-broker-qa.web.cern.ch",
             # Profiler reprot viewer dependencies
             "maxcdn.bootstrapcdn.com",
             "cdnjs.cloudflare.com",
             "ajax.googleapis.com",
+            # For CERN Matomo
+            "webanalytics.web.cern.ch",
         ],
     },
     'content_security_policy_report_only': False,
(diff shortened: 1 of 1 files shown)


invenio.cfg
```diff
@@ -550,3 +550,9 @@ OPENAIRE_DIRECT_INDEXING_ENABLED = True
 """Enable sending published records for direct indexing at OpenAIRE."""
 
 RDM_RECORDS_SERVICE_COMPONENTS = DefaultRecordsComponents + [OpenAIREComponent]
+
+MAIL_DEFAULT_SENDER = "Zenodo <info@zenodo.org>"
```
slint, line 554: @lnielsen what's the `noreply@cern.ch` address equivalent we should be using for Zenodo?

lnielsen: I'm checking with Emmanuel.

lnielsen: We can use ``noreply@zenodo.org``

invenio.cfg
```diff
@@ -567,4 +567,10 @@ USERS_RESOURCES_ADMINISTRATION_ENABLED = True
...
 COMMUNITIES_ADMINISTRATION_DISABLED = False
-"""Enable communities administration."""
\ No newline at end of file
+"""Enable communities administration."""
+
+MAIL_DEFAULT_SENDER = "Zenodo <info@zenodo.org>"
+"""Default e-mail address sender."
+
+REST_ENABLE_CORS = True
+"""Enable CORS validation.""""
```
zzacharo, line 576: Should we evaluate the [CSP rules](https://github.com/zenodo/zenodo-rdm/pull/542/files#diff-35ea7d0c9f0d3771b33ec0ac671404047bdff444801fcf084c39437c21107ebfR85-R86) as well? Especially, the fact that we allow `unsafe-inline` and `unsafe-eval`?

invenio.cfg
alejandromumo: Modified the config for Zenodo-RDM 👍

Issue #544, opened 2023-09-25
OpenAIRE: enable delete / restore 


alejandromumo:
After a record is deleted / restored, the platform should request OpenAIRE to delete / index the record on their side. 

The OpenAIRE backend was migrated (https://github.com/zenodo/zenodo-rdm/pull/515) and already supports both operations.

The record service [component](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/components/record_deletion.py) triggers a `delete_record` and `restore_record` calls. We can hook zenodo-rdm OpenAIRE component to these new calls.

PR #548, opened 2023-09-26
openaire: added delete and restore components

alejandromumo:
closes https://github.com/zenodo/zenodo-rdm/issues/544


site/tests/openaire/test_components.py
```diff
@@ -38,6 +41,107 @@ def test_on_publish(
...
+            data=json.dumps(
+                {
+                    "originalId": f"10.5281/zenodo.{recid}",
+                    "collectedFromId": "opendoar____::2659",
+                }
+            ),
+        )
+    ]
+
+    mocked_session.post.assert_has_calls(post_calls)
```
kpsherva, line 146: does it make sense to check the returns and arguments too ?

alejandromumo: Returns are completely [mocked](https://github.com/alejandromumo/zenodo-rdm/blob/add_openaire_delete_restore/site/tests/openaire/conftest.py#L29) since we're not hitting any real API. 

Arguments are already being checked since we're passing specific `call` objects: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_has_calls

zzacharo (approved):

PR #558, opened 2023-10-03
metrics: omit metric and log exception on failure

### LINKED ISSUE #554: metrics: backup value for metric evaluation failure
Currently, all metrics fail to show even if one fails. We want to send a nil/None/0 (TBD) value for a metric when the evaluation fails for any reason. This value should be prometheus friendly, meaning that it shouldn’t cause a sudden drop/spike in the graphs.

yashlamba:
Fixes #554 

We heard back from Marek that there's no good value we can return for Promtheus, so we are omitting the metric itself and logging an exception so we can keep track through sentry.

### DIFF
--- a/site/zenodo_rdm/metrics/utils.py
+++ b/site/zenodo_rdm/metrics/utils.py
@@ -26,8 +26,8 @@ def calculate_metrics(metric_id, cache=True):
     for metric in result:
         try:
             metric["value"] = metric["value"]()
-        except:
-            current_app.logger.exception(f"Metric evaluation failed {metric['name']}")
+        except Exception:
+            current_app.logger.exception("Metric evaluation failed", extra={"metric": metric['name']})
 
     if cache:
         current_cache.set(
--- a/site/zenodo_rdm/metrics/utils.py
+++ b/site/zenodo_rdm/metrics/utils.py
@@ -27,7 +27,9 @@ def calculate_metrics(metric_id, cache=True):
         try:
             metric["value"] = metric["value"]()
         except Exception:
-            current_app.logger.exception("Metric evaluation failed", extra={"metric": metric['name']})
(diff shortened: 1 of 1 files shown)


site/zenodo_rdm/metrics/utils.py
```diff
@@ -24,7 +24,10 @@ def calculate_metrics(metric_id, cache=True):
     result = deepcopy(current_app.config["METRICS_DATA"][metric_id])
 
     for metric in result:
-        metric["value"] = metric["value"]()
+        try:
+            metric["value"] = metric["value"]()
+        except:
```
slint, line 29:
```suggestion
        except Exception:
```

minor: it's better to specify `Exception` since we don't want e.g. to handle a `SysExit` or `KeyboardInterrupt` (see e.g. [this article](https://www.30secondsofcode.org/python/s/bare-except/))

site/zenodo_rdm/metrics/utils.py
```diff
@@ -24,7 +24,10 @@ def calculate_metrics(metric_id, cache=True):
...
+            current_app.logger.exception(f"Metric evaluation failed {metric['name']}")
```
slint, line 30:
```suggestion
            current_app.logger.exception(f"Metric evaluation failed: %s", metric['name'])
```

...or...

```suggestion
            current_app.logger.exception("Metric evaluation failed", extra={"metric": metric['name']})
```

minor: this helps [Sentry to group](https://docs.sentry.io/platforms/python/usage/sdk-fingerprinting/) errors better (otherwise the exception message changes everytime). I know that we have only one metric type, but it's good to have it right in case someone is looking on how to log an exception and copy-pastes this :)

slint (approved):

PR #574, opened 2023-10-11
test workflow

zzacharo:


site/zenodo_rdm/decorators.py
```diff
@@ -25,9 +27,17 @@ def cached_unless_authenticated_or_flashes(timeout=50, key_prefix="default"):
     def caching(f):
         @wraps(f)
         def wrapper(*args, **kwargs):
+            # we compute the cache key based on the `Dockerfile.IMAGE_BUILD_TIMESTAMP`
+            # env variable if set or we generate a new timestamp every time
+            # i.e no cache. This happens in the case of local development
+            cache_prefix = (
+                f"{key_prefix}_{current_app.config.get('IMAGE_BUILD_TIMESTAMP')}"
+                if current_app.config.get("IMAGE_BUILD_TIMESTAMP") is not None
+                else f"{key_prefix}_{datetime.datetime.now():%Y%m%d_%H%M%S}"
```
slint, line 36:
```suggestion
                else key_prefix
```

This is a bit dangerous since you might end up adding a lot of big keys in Redis accidentally.

For local development I would say we can use `app.debug` or similar to disable caching.

PR #580, opened 2023-10-12
files: offload download

ntarocco:
Requires: https://gitlab.cern.ch/zenodo/zenodo-rdm-openshift/-/merge_requests/19

### DIFF
--- a/site/zenodo_rdm/files.py
+++ b/site/zenodo_rdm/files.py
@@ -1,0 +1,101 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 CERN.
+#
+# ZenodoRDM is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Zenodo files utilities."""
+
+import mimetypes
+from urllib.parse import urlsplit, urlunsplit
+
+import requests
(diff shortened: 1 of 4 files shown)


site/zenodo_rdm/files.py
```diff
@@ -0,0 +1,92 @@
...
+        host = current_app.config["ZENODO_EOS_OFFLOAD_HTTPHOST"]
+        redirect_base_path = current_app.config["ZENODO_EOS_OFFLOAD_REDIRECT_BASE_PATH"]
+        base_path = urlsplit(self.fileurl).path
+        eos_resp = requests.get(
+            f"{host}/{base_path}",
+            auth=HTTPKerberosAuth(DISABLED),
+            verify=False,
+            allow_redirects=False,
+        )
+        assert eos_resp.status_code == 307
```
slint, line 43:
```suggestion
        if eos_resp.status_code != 307:
            current_app.logger.error("EOS redirect failed")
            return
```

Maybe we return and fallback to serving through the app? I know it's a bit risky since if it starts happening permanently we won't notice (that's why I also proposed the error logging).

slint (approved):
Nice, thanks a lot! One comment about the fallback (to avoid completely shutting files off in case there's an issue with the EOS redirect).

PR #597, opened 2023-10-14
config: fix orcid access token url.

alejandromumo:
closes https://github.com/zenodo/rdm-project/issues/331


slint (approved):
Confirmed against legacy prod config, and it's correct

PR #599, opened 2023-10-14
files: better exception visibility

kpsherva:
Guys, could we get this in first thing on the next release? 
I would like to make this exception a bit less vague https://zenodo-sentry.web.cern.ch/sentry/zenodo-rdm/issues/29446800/?query=is%3Aunresolved%20release%3A2.3.5


lnielsen (approved):
Approved assuming variables are always set

PR #602, opened 2023-10-16
config: re-enable openaire oauth app.

alejandromumo:
closes https://github.com/zenodo/rdm-project/issues/371

@slint it seems to me that the client ID was updated on OpenAIRE side and the issue should be fixed ™️ , however can only test it when we deploy.


slint:
Speed-merged since it's just config change :)

PR #608, opened 2023-10-16
legacy: fix files API querying by id

zzacharo:


site/zenodo_rdm/legacy/serializers/schemas/zenodojson.py
```diff
@@ -211,3 +194,25 @@ def dump_status(self, obj):
         return "published"
 
     stats = fields.Nested(StatsSchema)
+
+    def dump_files(self, obj):
+        """Dump files."""
```
zzacharo, line 199: @slint if we are fine with this then we can move it legacy common.

site/zenodo_rdm/legacy/services.py
```diff
@@ -331,18 +332,33 @@ def get_record_by_bucket_id(self, bucket_id):
...
-            .join(FileInstance, ObjectVersion.file_id == FileInstance.id)
-            .join(RDMDraft, ObjectVersion.bucket_id == RDMDraft.bucket_id)
-            .join(PersistentIdentifier, PersistentIdentifier.object_uuid == RDMDraft.id)
-            .filter(
-                FileInstance.id == file_id,
-                PersistentIdentifier.pid_type == "recid",
-                PersistentIdentifier.pid_value == pid_value,
-                PersistentIdentifier.object_type == "rec",
+
+        def query_file_id_by_model_cls(model_cls):
```
zzacharo, line 336: I think this is used only for drafts...I am not sure if having this fallback makes sense

site/zenodo_rdm/legacy/serializers/schemas/legacyjson.py
```diff
@@ -139,9 +139,8 @@ def dump_files(self, obj):
         bucket_url = obj["links"].get("bucket")
         files_url = obj["links"].get("files")
         for key, f in obj["files"].get("entries", {}).items():
-            file_id = f["id"]
             if files_url:
-                links = {"self": f"{files_url}/{file_id}"}
+                links = {"self": f"{files_url}/{key}"}
```
zzacharo, line 143: that was producing wrong link i.e `api/records/<recid>/files/<file_id>`

slint: We have to be careful, some of these links are important for partners (like LORY).

zzacharo: if they had hardcode bucket url, I checked and it still works but the generated links changed to working ones.

site/zenodo_rdm/legacy/serializers/schemas/legacyjson.py
```diff
@@ -136,14 +136,12 @@ def dump_owner(self, obj):
...
         files_url = obj["links"].get("files")
         for key, f in obj["files"].get("entries", {}).items():
-            file_id = f["id"]
             if files_url:
-                links = {"self": f"{files_url}/{file_id}"}
-            if bucket_url:
-                links["download"] = f"{bucket_url}/{key}"
+                self_html = obj["links"]["self_html"]
+                links = {"self": f"{files_url}/{key}"}
+                links["download"] = f"{self_html}/files/{key}"
```
zzacharo, line 144: @lnielsen suggested to put here the html file download link as it is more likely to keep consistent.

zzacharo: DIscussed with @slint: we cannot do this because then other UIs fetching files from that link will hit CORS issues.

zzacharo:
superseded by https://github.com/zenodo/zenodo-rdm/pull/618

PR #619, opened 2023-10-19
sort: add legacy sort options

zzacharo:


invenio.cfg
```diff
@@ -302,6 +302,68 @@ ZENODO_LEGACY_SEARCH_MAP = {
     "size": "files.totalbytes",
 }
 
+LEGACY_SORT_OPTIONS = {
+    "-bestmatch": dict(
+        title=_("Best match [Least relevant]"),
```
slint, line 307: comment/question: do you know why we use square brackets for the sorting order description? I feel that regular parentheses would look better in this case.

PR #622, opened 2023-10-20
Add legacy search params

zzacharo:


site/zenodo_rdm/params.py
```diff
@@ -0,0 +1,103 @@
...
+    all_versions = fields.Boolean()
+    communities = fields.List(fields.String())
+    status = fields.String()
+    type = fields.String()
+    subtype = fields.String()
+
+    @post_load
+    def dump_all_versions(self, data, **kwargs):
+        if "all_versions" in data:
+            data["allversions"] = data["all_versions"]
```
slint, line 101: minor/question: does this take into account when just passing `?all_versions` (i.e. without any value)?

slint (approved):
LGTM, one comment on `?all_versions` behaviour

site/zenodo_rdm/params.py
```diff
@@ -0,0 +1,107 @@
...
+    all_versions = fields.Boolean()
+    communities = fields.List(fields.String())
+    status = fields.String()
+    type = fields.String()
+    subtype = fields.String()
+
+    @pre_load
+    def load_all_versions(self, data, **kwargs):
+        if "all_versions" in data:
+            if not data["all_versions"]:
```
zzacharo, line 101: @slint now it is taken into account

slint, line 101:
```suggestion
            if data["all_versions"] in (None, ""):
```

Maybe this is more safe against `False` or `0`, (or whatever Flask considers Falsy querystring params)?

PR #625, opened 2023-10-23
github: added default repo creators.

alejandromumo:
closes https://github.com/zenodo/rdm-project/issues/434

### DIFF
--- a/site/zenodo_rdm/github/release.py
+++ b/site/zenodo_rdm/github/release.py
@@ -79,4 +79,24 @@ class ZenodoGithubRelease(RDMGithubRelease):
             if citation_metadata:
                 output.update(citation_metadata)
 
+        if not output.get("creators"):
+            # Get owner from Github API
+            owner = self.get_owner()
+            if owner:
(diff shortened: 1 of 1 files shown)


site/zenodo_rdm/github/release.py
```diff
@@ -79,4 +79,24 @@ def metadata(self):
             if citation_metadata:
                 output.update(citation_metadata)
 
+        if not output.get("creators"):
```
alejandromumo, line 82: This "pollutes" the release API a bit, I would expect that all the metadata-related operations would be in the metadata class `RDMReleaseMetadata` (or `ZenodoReleaseMetadata`). 

However, that requires that we rework the code and we don't want to do it right now I suppose.

alejandromumo, line 82: This is duplicated from `invenio-rdm-records` but the way we structured the code forces this. We'd need a rework on how the metadata is computed, maybe something for later.

slint (approved):

PR #636, opened 2023-10-26
files: normalize headers encoding

kpsherva:
* fix attachment download flag
* closes https://github.com/zenodo/rdm-project/issues/401

### DIFF
--- a/site/zenodo_rdm/files.py
+++ b/site/zenodo_rdm/files.py
@@ -8,12 +8,14 @@
 """Zenodo files utilities."""
 
 import mimetypes
+import unicodedata
 from urllib.parse import urlsplit, urlunsplit
 
 import requests
 from flask import current_app, make_response, request
 from invenio_files_rest.helpers import sanitize_mimetype
 from invenio_files_rest.storage.pyfs import pyfs_storage_factory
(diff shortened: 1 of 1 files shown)


site/zenodo_rdm/files.py
```diff
@@ -55,6 +57,7 @@ def _get_eos_redirect_path(self):
     def send_file(self, filename, **kwargs):
```
slint, line 57:
```suggestion
    def send_file(self, filename, as_attachment=False, **kwargs):
```

minor: can't we expose like this actually instead of cherry-picking `kwargs`? Actually I would expose all params from [the original `send_file`](https://github.com/inveniosoftware/invenio-files-rest/blob/74d89cc6cefdab267eacbe8bb0f8f37c2c715590/invenio_files_rest/storage/base.py#L103-L112), so that we don't forget to handle some of them.

slint (approved):
LGTM, one minor comment on adding all the params to the method.

site/zenodo_rdm/files.py
kpsherva: I assumed there was a reason of not implementing them initially, but good point, I will change it, thanks

PR #646, opened 2023-11-02
landing page: Add legacy subjects to additional details

jennur:
Closes https://github.com/zenodo/rdm-project/issues/446

Needs: https://github.com/inveniosoftware/invenio-app-rdm/pull/2520 << Screenshots and explanation

### DIFF
--- a/templates/semantic-ui/zenodo_rdm/records/detail.html
+++ b/templates/semantic-ui/zenodo_rdm/records/detail.html
@@ -69,6 +69,14 @@
 {%- endblock record_files -%}
 
 
+{# Additional details #}
+{%- block record_details -%}
+  <section id="additional-details" class="rel-mt-2" aria-label="{{ _('Additional record details') }}">
+    {%- include "zenodo_rdm/records/details/details.html" %}
+  </section>
+{%- endblock record_details -%}
+
+
 {%- block record_footer -%}
   {% if citations_enabled and not is_preview %}
     <section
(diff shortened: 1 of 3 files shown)


templates/semantic-ui/zenodo_rdm/records/detail.html
```diff
@@ -69,6 +69,14 @@ <h4 class="inline">{{ record.ui.access_status.title_l10n }}</h4>
 {%- endblock record_files -%}
 
 
+{# Additional details #}
+{%- block record_details -%}
```
kpsherva, line 73: doesn't it completely override the whole block in the invenio_app_rdm/records/detail.html ? was that intentional? because I see there is quite some information displayed, and I think it will be ommitted by this change. See https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/details/details.html
if I am wrong, please ignore the comment

jennur: Yes, and no. It replaces the app-rdm template with the zenodo-rdm template I added here, which is extending the app-rdm template and adds the extra content in the new `extras` block I added in the app-rdm PR https://github.com/inveniosoftware/invenio-app-rdm/pull/2520, so all the content is still there.

kpsherva: I think it can stay like this for now, I am a bit confused with the hierarchy but I think it is because originally organised this way. 
Thanks for checking and testing!

kpsherva (approved):

Issue #421, opened 2023-07-26
Citations box: remove blue no-results box


jennur:
Replace the blue box with a simple note.

PR #650, opened 2023-11-08
ui: adding css for rendered editor content

0einstein0:
Partially Closes Issue [#421](https://github.com/zenodo/rdm-project/issues/421)
Screenshot 2023-11-08 at 15 09 45

### DIFF
--- a/assets/less/zenodo-rdm/globals/site.overrides
+++ b/assets/less/zenodo-rdm/globals/site.overrides
@@ -22,6 +22,19 @@ body {
   }
 }
 
+.mce-content {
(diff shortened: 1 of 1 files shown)


assets/less/zenodo-rdm/globals/site.overrides
```diff
@@ -22,6 +22,19 @@ body {
   }
 }
 
+.mce-content {
```
kpsherva, line 25: I would move this to invenio-theme instead, since it affecting the whole core, not only zenodo

PR #651, opened 2023-11-09
versions: display only active versions

anikachurilova:
* closes https://github.com/zenodo/rdm-project/issues/515


ptamarit (approved):
Remark: this seems to not be used in Zenodo (the app-rdm template is used instead).

anikachurilova:
> Remark: this seems to not be used in Zenodo (the app-rdm template is used instead).

@ptamarit seems like here it is using the template from zenodo: https://github.com/zenodo/zenodo-rdm/blob/master/templates/themes/horizon/invenio_communities/details/home/index.html#L11

PR #665, opened 2023-11-29
schema: updated json serializer list schema

0einstein0:
Closes Issue https://github.com/zenodo/rdm-project/issues/426

Screenshot 2024-01-23 at 10 38 30

### DIFF
--- a/site/zenodo_rdm/legacy/serializers/__init__.py
+++ b/site/zenodo_rdm/legacy/serializers/__init__.py
@@ -8,7 +8,7 @@
 """Zenodo legacy serializers."""
 
 from flask_resources import BaseListSchema, JSONSerializer, MarshmallowSerializer
-from marshmallow import fields, post_dump
+from marshmallow import fields, missing, post_dump
 
 from .schemas import (
     LegacyFileListSchema,
@@ -33,6 +33,12 @@ class LegacyListSchema(BaseListSchema):
         return data.get("hits", {}).get("hits", [])
 
 
+class ZenodoListSchema(BaseListSchema):
+    """Zenodo top-level List schema."""
+
+    sortBy = fields.Field(load_only=True)
+
+
 class LegacyJSONSerializer(MarshmallowSerializer):
     """Legacy metadata serializer."""
 
@@ -53,7 +59,7 @@ class ZenodoJSONSerializer(MarshmallowSerializer):
         super().__init__(
             format_serializer_cls=JSONSerializer,
             object_schema_cls=ZenodoSchema,
-            list_schema_cls=BaseListSchema,
+            list_schema_cls=ZenodoListSchema,
         )
 
 


slint (commented):
LGTM, one last part about translating legacy options

site/zenodo_rdm/legacy/serializers/__init__.py
```diff
@@ -33,6 +33,27 @@ def unwrap_hits(self, data, many, **kwargs):
         return data.get("hits", {}).get("hits", [])
 
 
+class ZenodoListSchema(BaseListSchema):
+    """Legacy top-level array/list schema for JSON Serializer."""
+
+    links = fields.Method("get_links")
+    sortBy = fields.Method("get_sorting_option")
```
lnielsen, line 40: I've checked and ``sortBy`` was not included in the legacy JSON, but since we'll be adding in ``BaseListSchema`` we'll need to set it to None here.

```suggestion
    sortBy = None
```

site/zenodo_rdm/legacy/serializers/__init__.py
```diff
@@ -33,6 +33,27 @@ def unwrap_hits(self, data, many, **kwargs):
         return data.get("hits", {}).get("hits", [])
 
 
+class ZenodoListSchema(BaseListSchema):
+    """Legacy top-level array/list schema for JSON Serializer."""
+
+    links = fields.Method("get_links")
```
lnielsen, line 39: The method as it is should be added in ``BaseListSchema`` in flask-resources because it affects all serializers using the MarshmallowSerializer.

Here we should still override the method being added to ``BaseListSchmea`` because it will generate links like:

- ``https://zenodo.org/api/records?q=&sort=newest&page=1&size=10``

Which in legacy is:
-  ``https://zenodo.org/api/records?q=&sort=mostrecent&page=1&size=10

site/zenodo_rdm/legacy/serializers/__init__.py
```diff
@@ -33,6 +33,12 @@ def unwrap_hits(self, data, many, **kwargs):
         return data.get("hits", {}).get("hits", [])
 
 
+class ZenodoListSchema(BaseListSchema):
```
ptamarit, line 36: nit: Seeing the class below prefixed `Legacy`, and this one not prefixed, made me think that `ZenodoListSchema` was related to the modern API (until I realized that the package name made all this code legacy).
Not needed to be changed since other existing classes like `ZenodoSchema` are not prefixed neither.

slint: Agree that the naming is not consistent throughout the legacy module (because of a code move/rename), and there's also some special naming/meaning around the terms "Legacy" and "ZenodoJSON" (the latter is the case for this class).

There's a bit of a counter-incentive also on putting effort into polishing the legacy layer since it's "throwaway" code that will be pruned once we deprecate the legacy REST API.

site/zenodo_rdm/legacy/serializers/__init__.py
```diff
@@ -33,6 +33,12 @@ def unwrap_hits(self, data, many, **kwargs):
...
+    """Zenodo top-level List schema."""
+
+    sortBy = fields.Field(load_only=True)
```
slint, line 39: question: do you know where the `sort` parameter is loaded/dumped now?

anikachurilova: question: why do we need to reimplement this here? it's already being done in `flask-resources`

0einstein0: > question: do you know where the `sort` parameter is loaded/dumped now?

@slint  by sort, do you mean the `sortBy` ? I didn't track it, thinking it might be being used somewhere as it is allowed everywhere so I set load_only=True, should I just override and return missing?

0einstein0: > question: why do we need to reimplement this here? it's already being done in `flask-resources`

@anikachurilova  this is in response to this comment https://github.com/zenodo/zenodo-rdm/pull/665#discussion_r1424300682

slint: > > question: do you know where the `sort` parameter is loaded/dumped now?
> 
> @slint by sort, do you mean the `sortBy` ? I didn't track it, thinking it might be being used somewhere as it is allowed everywhere so I set load_only=True, should I just override and return missing?

I mean if one passes now `GET /api/records?sort=mostrecent`, I would expect:

1. the parameter `sort=mostrecent` is passed down to the service to return records sorted by most recent first
2. in the response I see in the `links`, `self: /api/records?sort=mostrecent&page=1&size=25, ...`

I don't see anywhere in the schema a field interacting with the `?sort=...` parameter

PR #674, opened 2023-12-08
landing page: add blr related works section

jennur:
Closes https://github.com/zenodo/rdm-project/issues/567

Needs https://github.com/inveniosoftware/react-searchkit/pull/260

**Edit:** After discussing with Lars, it was decided to remove the search bar for now, as there are issues related to having the initial query string (to show only related results for the specific record) + additional query strings added by the user. In order to add the search bar back, these are the known alternatives that will need a solution:
- if the query string is added to the state (`queryString`), the string shows up in the input field. The input field should be empty, but the query string needs to remain.
- if the query string is added directly to the endpoint url parameters, the search does not work, as there will be two `q` parameters once another search is performed.
- the query string cannot be set as a filter, as this requires the implementation of custom filters in the backend.

### Screenshots
Screenshot 2023-12-13 at 8 23 23 AM
Screenshot 2023-12-13 at 8 23 33 AM
Screenshot 2023-12-13 at 8 26 09 AM

### DIFF
--- a/assets/less/zenodo-rdm/globals/site.overrides
+++ b/assets/less/zenodo-rdm/globals/site.overrides
@@ -139,3 +139,8 @@ a.inverted {
     font-size: 1.45rem;
   }
 }
+
+.image.thumbnail-image {
+  height: 150px;
+  object-fit: cover;
+}
--- a/site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/blr-related-works/BlrSearch.js
+++ b/site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/blr-related-works/BlrSearch.js
@@ -35,12 +35,13 @@ import {
 } from "react-searchkit";
 import { OverridableContext } from "react-overridable";
 import { apiConfig } from "./api/config";
-import { Segment, Container, Header } from "semantic-ui-react";
+import { Segment, Container } from "semantic-ui-react";
 import { ResultsGridLayout, ResultsListLayout } from "./components/ResultsLayout";
 import { RecordGridItem, RecordListItem } from "./components/RecordItem";
 import { FilterContainer, Filter, FilterValues } from "./components/Filter";
 import { LayoutSwitchButtons } from "./components/LayoutSwitchButtons";
 import { NoResults } from "./components/NoResults";
+import { BlrResultsLoader } from "./components/BlrResultsLoader";
 
 const blrSearchAppID = "blrSearch";
 
@@ -54,16 +55,17 @@ const overriddenComponents = {
   [`${blrSearchAppID}.BucketAggregationValues.element`]: FilterValues,
   [`${blrSearchAppID}.LayoutSwitcher.element`]: LayoutSwitchButtons,
   [`${blrSearchAppID}.EmptyResults.element`]: NoResults,
+  [`${blrSearchAppID}.ResultsLoader.element`]: BlrResultsLoader,
 };
 
-export const BlrSearch = ({ endpoint, recordDOI, resourceType }) => {
+export const BlrSearch = ({ endpoint, recordDOI, resourceType, blrId }) => {
   const relationType = (resourceType) =>
     resourceType === "Journal article" || resourceType === "Book chapter"
       ? "ispartof"
       : "haspart";
 
   const queryString = (relationType, identifier) =>
-    `metadata.related_identifiers.relation_type.id:${relationType} AND metadata.related_identifiers.identifier:"${identifier}"`;
+    `parent.communities.ids:${blrId} AND metadata.related_identifiers.relation_type.id:${relationType} AND metadata.related_identifiers.identifier:"${identifier}"`;
 
   const searchApi = new InvenioSearchApi(apiConfig(endpoint));
 
@@ -77,37 +79,34 @@ export const BlrSearch = ({ endpoint, recordDOI, resourceType }) => {
   };
 
   return (
-    <>
-      <Header as="h2">Linked records</Header>
-      <OverridableContext.Provider value={overriddenComponents}>
-        <ReactSearchKit
-          appName={blrSearchAppID}
-          searchApi={searchApi}
-          urlHandlerApi={{ enabled: false }}
-          initialQueryState={initialState}
-        >
-          <>
-            <div className="flex align-items-center justify-space-between">
-              <BucketAggregation
-                agg={{ field: "resource_type", aggName: "resource_type" }}
-              />
-              <LayoutSwitcher />
-            </div>
+    <OverridableContext.Provider value={overriddenComponents}>
+      <ReactSearchKit
+        appName={blrSearchAppID}
+        searchApi={searchApi}
+        urlHandlerApi={{ enabled: false }}
+        initialQueryState={initialState}
+      >
+        <>
+          <div className="flex align-items-center justify-space-between">
+            <BucketAggregation
+              agg={{ field: "resource_type", aggName: "resource_type" }}
+            />
+            <LayoutSwitcher />
+          </div>
 
-            <Segment>
-              <ResultsLoader>
-                <ResultsMultiLayout />
-                <Error />
-                <EmptyResults />
-              </ResultsLoader>
+          <Segment>
+            <ResultsLoader>
+              <ResultsMultiLayout />
+              <Error />
+              <EmptyResults />
               <Container align="center" className="rel-pt-1">
                 <Pagination options={{ size: "mini", showEllipsis: true }} />
               </Container>
-            </Segment>
-          </>
-        </ReactSearchKit>
-      </OverridableContext.Provider>
-    </>
+            </ResultsLoader>
+          </Segment>
+        </>
+      </ReactSearchKit>
+    </OverridableContext.Provider>
   );
 };
 
@@ -115,4 +114,5 @@ BlrSearch.propTypes = {
   endpoint: PropTypes.string.isRequired,
   recordDOI: PropTypes.string.isRequired,
   resourceType: PropTypes.string.isRequired,
+  blrId: PropTypes.string.isRequired,
 };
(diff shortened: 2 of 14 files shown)


assets/less/zenodo-rdm/globals/site.overrides
```diff
@@ -138,4 +138,9 @@ a.inverted {
   .ui.medium.header {
     font-size: 1.45rem;
   }
+}
+
+.image.thumbnail-image {
+  height: 150px;
```
kpsherva, line 144: why not using the sematic ui size prop ?

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/blr-related-works/components/BlrSearchBar.js
```diff
@@ -0,0 +1,79 @@
...
+//
+// In applying this license, CERN does not
+// waive the privileges and immunities granted to it by virtue of its status
+// as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+import React from "react";
+import { PropTypes } from "prop-types";
+import { Input, Icon } from "semantic-ui-react";
+
+export const BlrSearchBar = ({
```
kpsherva, line 27: is this component any different that the default search bar? maybe I am missing context but my first thought is this PR recreates the whole search app and duplicates a lot of code. We could possibly use overridable features like in other searches

assets/less/zenodo-rdm/globals/site.overrides
jennur: Because I want to set the height, not the width.

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/blr-related-works/components/BlrSearchBar.js
jennur: You're right, the search bar override is not needed I think. I think I can solve it with the props only. Thanks!

Do you mean the search app from invenio-search-ui?

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/blr-related-works/components/Filter.js
```diff
@@ -0,0 +1,87 @@
...
+
+FilterContainer.propTypes = {
+  agg: PropTypes.object.isRequired,
+  updateQueryFilters: PropTypes.func.isRequired,
+  containerCmp: PropTypes.node.isRequired,
+};
+
+export const Filter = ({ valuesCmp }) => {
+  return (
+    <Dropdown text="Filter by type" button>
```
jrcastro2, line 52: Comment: By using the `text` prop, the user will not see what type of filter is selected unless he opens the dropodwn. I find it more intuitive if once we have selected the filter it's value is displayed. 
```suggestion
    <Dropdown placeholder="Filter by type" button>
```

jrcastro2 (approved):
LGTM!
Just left a minor comment about the filter, the PR also need to be rebased there are conflicts.

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/blr-related-works/components/Filter.js
jennur: Thanks! I do agree, although this dropdown is a bit different since it works like the filters on the regular search (with checkboxes), so multiple options can be checked.

jrcastro2: Right! Good point! In that case I would just display some text mentioning how many filters are selected like "X filters selected" so that the user is aware that there are filters selected when the dropdown is closed. However this can be addressed in another issue, if we do so please create the issue!

jennur: Thanks! I will test it!

jennur:
### Update
- Added counter for selected filters

Screenshot 2023-12-21 at 2 58 32 PM

Screenshot 2023-12-21 at 2 58 26 PM

Screenshot 2023-12-21 at 2 58 20 PM

anikachurilova:
Update: added a placeholder for results loader

relrec

assets/less/zenodo-rdm/globals/site.overrides
```diff
@@ -138,4 +138,9 @@ a.inverted {
   .ui.medium.header {
     font-size: 1.45rem;
   }
+}
+
+.image.thumbnail-image {
```
ntarocco, line 143: Does this works well in mobile version?

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/blr-related-works/BlrSearch.js
```diff
@@ -0,0 +1,121 @@
...
+    sortBy: "bestmatch",
+    sortOrder: "asc",
+    page: 1,
+    size: 4,
+    layout: "list",
+  };
+
+  return (
+    <>
+      <Header as="h2">Linked records</Header>
```
ntarocco, line 83: Will Zenodo be available in multiple languages at some point? We might have to mark all strings as translatable.

ntarocco (commented):
If meant to be reusable, it should be moved to InvenioRDM (maybe later?) and naming should be changed to be more generic (BLR -> Linked records).

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/blr-related-works/BlrSearch.js
ntarocco, line 83: This could be maybe moved to Jinja (if easy) to avoid the page flickering.

anikachurilova: As far as I remember, it was a problematic thing to implement... but anyway that would be another task to go around and make all the strings translatable, for now doing that would be useless

anikachurilova:
Update: added a placeholder for results loader
gif cats

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/blr-related-works/components/RecordItem.js
```diff
@@ -0,0 +1,126 @@
...
+    title,
+    typeIcon,
+    link,
+  } = layoutProps(result);
+
+  return (
+    <Item>
+      <Image
+        size="tiny"
+        src={`${result.links.self_html}/thumb250`}
```
slint, line 90: shelve: we should actually have this link generated in the service so that we don't have to manually construct it here.

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/blr-related-works/index.js
```diff
@@ -0,0 +1,52 @@
...
+const recordDOI = JSON.parse(blrContainer.dataset.recordDoi);
+const resourceType = JSON.parse(blrContainer.dataset.resourceType);
+const blrId = JSON.parse(blrContainer.dataset.blrId);
+
+const validTypes = [
+  "Figure",
+  "Taxonomic treatment",
+  "Book chapter",
+  "Journal article",
+  "Drawing",
```
slint, line 38: nit/shelve: we should be using the ID versions of these resource types (i.e. `image-figure`, `publication-taxonomictreatment`, etc.). But this can be done when we generalize the feature.

templates/semantic-ui/zenodo_rdm/records/detail.html
```diff
@@ -64,6 +64,33 @@ <h4 class="inline">{{ record.ui.access_status.title_l10n }}</h4>
 
 {# Additional details #}
 {%- block record_details -%}
+
+  {% if record.expanded and record.expanded.parent.communities.default.slug == "biosyslit" %}
```
slint, line 68: minor: it would be nice if there was a way to check this very nested attribute access (e.g. perhaps JS optional chaining) to avoid Jinja blowing up the page with an exception.

I would also extract this into a variable at the top of the template (to re-use below in the `javascript` block)

slint (approved):
Looks good overall, some minor comments and/or things to shelve.

Sign-off: @carlinmack @slint

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/blr-related-works/BlrSearch.js
```diff
@@ -0,0 +1,118 @@
...
+import { ResultsGridLayout, ResultsListLayout } from "./components/ResultsLayout";
+import { RecordGridItem, RecordListItem } from "./components/RecordItem";
+import { FilterContainer, Filter, FilterValues } from "./components/Filter";
+import { LayoutSwitchButtons } from "./components/LayoutSwitchButtons";
+import { NoResults } from "./components/NoResults";
+import { BlrResultsLoader } from "./components/BlrResultsLoader";
+
+const blrSearchAppID = "blrSearch";
+
+const overriddenComponents = {
```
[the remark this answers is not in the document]
anikachurilova: As far as I know there wasn't any discussion about results count. The amount of each type is displayed on the filters tho:
Screenshot 2024-01-24 at 11 04 40

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/blr-related-works/components/RecordItem.js
slint: Referenced in https://github.com/zenodo/zenodo-rdm/issues/682

slint (approved):

PR #678, opened 2023-12-14
citations: fixed html rendering.

alejandromumo:
Citations box is not rendering, throws an exception in the console due to malformed JSON. 

image

Which in turns points to:  https://github.com/zenodo/zenodo-rdm/blob/master/site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/citations/index.js#L36


jrcastro2 (approved):
To me both should be the same, if it fixes the issue LGTM

PR #684, opened 2024-01-30
http headers: use and adjust vnd.inveniordm.v1+json http accept header

anikachurilova:
* closes https://github.com/zenodo/rdm-project/issues/598

### DIFF
--- a/assets/js/components/landing_page/overrides/UpgradeLegacyRecordButton.js
+++ b/assets/js/components/landing_page/overrides/UpgradeLegacyRecordButton.js
@@ -29,11 +29,7 @@ export class UpgradeLegacyRecordButton extends Component {
     const { record } = this.props;
     try {
       const cancellableFetch = withCancel(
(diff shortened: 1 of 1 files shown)


site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
```diff
@@ -16,7 +16,6 @@ import FileUploader from "./FileUploader";
 
 const requestConfig = {
   headers: {
-    "Accept": "application/json",
     "Content-Type": "multipart/form-data",
```
kpsherva, line 19: I think it needs to specify the accept header, it looks to me that here https://github.com/zenodo/zenodo-rdm/pull/684/files#diff-e1f7092b13856cf547fcc0930acf3af64025c718e6860dd6e619c528302d60cbL85 headers are overridden, not updated

zzacharo (approved):

PR #708, opened 2024-02-07
global: updated robots.txt

alejandromumo:
closes https://github.com/zenodo/ops/issues/358

 * added rule to allow crawling of /api/files

### DIFF
--- a/static/robots.txt
+++ b/static/robots.txt
@@ -8,4 +8,5 @@ Disallow: /me
 Disallow: /login
 Disallow: /logout
 Disallow: /records/*/preview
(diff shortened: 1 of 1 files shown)


ptamarit (approved):
Looking at the [order of precedence for rules](https://developers.google.com/search/docs/crawling-indexing/robots/robots_txt#order-of-precedence-for-rules), I think that the following will apply:

> crawlers use the most specific rule based on the length of the rule path

Which matches with the behavior we want to achieve (allowing `/api/records/*/files` and disallowing the rest of `/api`).

Issue #732, opened 2024-02-20
Import EuroSci Vocabulary 


alejandromumo:
The vocabulary can be fetched from  [here](https://op.europa.eu/en/web/eu-vocabularies/dataset/-/resource?uri=http://publications.europa.eu/resource/dataset/euroscivoc)

PR #737, opened 2024-02-22
vocabularies: added European Science Vocbulary

alejandromumo:
closes https://github.com/zenodo/zenodo-rdm/issues/732


slint (approved):
LGTU, with @yashlamba

PR #752, opened 2024-02-27
scripts: added script to move records and communities in bulk.

alejandromumo:
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1677

### DIFF
--- a/scripts/ec_transfer_records.py
+++ b/scripts/ec_transfer_records.py
@@ -1,0 +1,142 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2024 CERN.
+#
+# Zenodo-RDM is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+"""Script to transfer communities and records to new parent.
+
+Usage:
+
+.. code-block:: python
+
+    _to = "ec_community"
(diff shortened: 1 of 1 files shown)


scripts/ec_transfer_records.py
```diff
@@ -0,0 +1,142 @@
...
+from invenio_rdm_records.proxies import current_rdm_records
+from invenio_records_resources.services.uow import UnitOfWork
+from invenio_search.engine import dsl
+from werkzeug.local import LocalProxy
+
+community_service = LocalProxy(lambda: current_communities.service)
+records_communities_service = LocalProxy(
+    lambda: current_rdm_records.record_communities_service
+)
+records_service = LocalProxy(lambda: current_rdm_records.records_service)
```
jrcastro2, line 35: Question: Do you need the `LocalProxy`? it's already a proxy (We might miss something ....)

jrcastro2 (commented):
LGTU! Only one comment about the services

scripts/ec_transfer_records.py
alejandromumo: Without using `LocalProxy`, the value will be resolved as soon as the file is loaded and, in that moment, the app context from flask may not exist yet. 

This way, the value will be resolved on demand (when the app context already exists theoretically).

PR #766, opened 2024-03-06
Add cookie banner

### LINKED ISSUE #759: Cookie policy banner
Implement a cookie consent banner on zenodo.org and all subdomains (about.zenodo.org, ...)

- [x] On every page (i.e. implement in base template somehow) there must be a small banner in the bottom (something like the banner screenshot below). Saying: ``(i) This website uses cookies. Find out more on how we use cookies.      Accept all cookies | Accept only essential cookies | X``
    - [x] The word ``cookies`` should be linked to a configurable URL (e.g. ``https://about.zenodo.org/cookie-policy``)
    - [ ] Allow instances to override the banner (by overriding the template), and to configure if they want it or not.
    - [ ] A global template context processor should inject if the cookie consent cookie is set to true/false.

Image

- JS for cookie banner:
  - [x] On pressing Accept all, JS should be used to set a cookie (``cookie_consent``). The consent cookie must work across all Zenodo subdomains.
  - [x] The JS code in addition must unset ``_pk_...`` cookies on every request in case the cookie consent is not set or if it is set to false (this way we can  allow users to withdraw consent from analytics cookies and it will work across all Zenodo subdomains).

- JS for Matomo
  - [x] The Matomo tracking code should do the following (see https://developer.matomo.org/guides/tracking-consent):

```
// User require user to accept Matomo cookies before setting them 
_paq.push(['requireCookieConsent']);
{% if cookie_consent %}
// User accepted Matomo cookies
_paq.push(['setCookieConsentGiven']);
{% endif %}
```

- [x] Add cookie policy menu item in bottom right of footer (https://github.com/zenodo/zenodo-docs-user/commit/96beb4d8a8f11ae5bad1626db5de776adbd26125)
image

- [x] Ensure it looks good on all screen sizes
- [x] All static sites must implement same cookie banner. https://github.com/zenodo/zenodo-docs-user/pull/115
- [ ] Cookie policy page must have the following buttons:
  - "Withdraw consent" if cookie consent is set to true
  - "Accept all cookies" if cookie consent is set to false or not set
  - "Accept only essential cookies" if cookie consent is not set

carlinmack:
Close #759 

Check above issue for tasks which still need to be done.

#### Behaviour

* By default the cookie banner is hidden
* JS checks whether a `cookie_consent` cookie is set.
* If not the banner is displayed
* If it is set and it is set to "all", the Matomo code is executed

* If the user presses any input on the banner, the banner is hidden. The all button will additionally call the Matomo code
* The cookie is set with expiry of one year and will work across subdomains as `SameSite=None; Secure`. The secure attributes means that this behaviour will only work with HTTPS

image


templates/semantic-ui/zenodo_rdm/trackingcode/trackingcode-prod.html
```diff
@@ -1,15 +1,79 @@
...
+      setCookie("cookie_consent","essential");
+    });
+
+  $('#cookies-all')
+    .on('click', function () {
+      $(this)
+        .closest('.message')
+        .transition('fade');
+      setCookie("cookie_consent","all");
+      _paq.push(['rememberCookieConsentGiven']);
```
ptamarit, line 61: **non-blocking:** Since Matomo is setting cookies based in its response based on the `remember` and `forget` parameters, I think that we could rely on their cookies instead of setting setting our own `cookie_consent`.

assets/less/zenodo-rdm/globals/site.overrides
```diff
@@ -143,4 +143,24 @@ a.inverted {
...
+    align-items: center;
+
+    .buttons {
+      display: flex;
+    }
+}
+
+.cookie-banner.hidden {
+    display: none !important;
+}
```
ptamarit, line 166: Keeping 2-spaces indentation.

```suggestion
.cookie-banner {
  position: fixed !important;
  left: 0;
  right: 0;
  bottom: 0;
  z-index: 1;
  padding: 0.5em 1.75em 0.5em 1em !important;
  display: flex !important;
  justify-content: space-between;
  align-items: center;

  .buttons {
    display: flex;
  }
}

.cookie-banner.hidden {
  display: none !important;
}
```

ptamarit (approved):

ptamarit:
**non-blocking**: not sure of the added value of having 10 commits in this case. I would personally squash them to 1.

invenio.cfg
```diff
@@ -201,6 +201,7 @@ THEME_FOOTER_TEMPLATE = 'zenodo_rdm/footer.html'
 # Enable the below on the dedicated environment
 # THEME_TRACKINGCODE_TEMPLATE = "zenodo_rdm/trackingcode/trackingcode-qa.html"
 
+COOKIE_POLICY = "https://about.zenodo.org/cookie-policy"
```
alejandromumo, line 204:
```suggestion
COOKIE_POLICY_URL = "https://about.zenodo.org/cookie-policy"
```

nit: naming wise, I'd make it more explicit that it's a URL. To avoid clashes In case we want to add a flag to enable or disable the banner

PR #793, opened 2024-03-19
support: fix user creation set unique login

ptamarit:
Fixes: https://github.com/zenodo/ops/issues/389

Creating users in Zammad fails with the error "This object already exists.".
There's a unique constraint on the `login` column, and we do not set the `login` when creating users.
This pull requests sets the `login` to the same value as the `email`, as is the case of the already created users.


ptamarit:
Turns out the problem was a problem with desynchronized sequences, and not with missing login.

PR #795, opened 2024-03-19
OpenAIRE direct indexing license fix

slint:
- **scripts: allow filtering deployments**
- **openaire: fix license serialization**

### DIFF
--- a/site/zenodo_rdm/openaire/serializers/schema.py
+++ b/site/zenodo_rdm/openaire/serializers/schema.py
@@ -156,7 +156,7 @@ class OpenAIRESchema(Schema):
         public_record = access["record"] == "public"
         public_files = access["files"] == "public"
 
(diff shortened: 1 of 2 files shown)


site/zenodo_rdm/openaire/serializers/schema.py
```diff
@@ -156,7 +156,7 @@ def get_license_code(self, obj):
         public_record = access["record"] == "public"
         public_files = access["files"] == "public"
 
-        access_settings = obj.get("parent").get("access").get("settings")
+        access_settings = obj.get("parent").get("access", {}).get("settings", {})
```
alejandromumo, line 159:
```suggestion
        access_settings = obj.get("parent", {}).get("access", {}).get("settings", {})
```

Just for those older records we might have around

alejandromumo (approved):

PR #798, opened 2024-03-21
bug: don't call uptime robot when no API key passed

carlinmack:
Close https://github.com/zenodo/ops/issues/392

Error: 

```bash
[2024-03-20 16:12:19,477: ERROR/ForkPoolWorker-17] Metric evaluation failed
Traceback (most recent call last):
  File "/Users/cmackenz/git/zenodo-rdm/site/zenodo_rdm/metrics/utils.py", line 29, in calculate_metrics
    metric["value"] = metric["value"]()
  File "/Users/cmackenz/git/zenodo-rdm/site/zenodo_rdm/metrics/api.py", line 107, in get_uptime
    for d in res.json()["monitors"]
KeyError: 'monitors'
```

The key is not found as when run locally the API returns:

```json
{"stat": "fail", "error": {"type": "missing_parameter", "parameter_name": "api_key", "message": "api_key parameter is missing."}}
```

Also `datetime.utcnow()` is deprecated so I have exchanged it for the correct implementation https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow

### DIFF
--- a/site/zenodo_rdm/metrics/api.py
+++ b/site/zenodo_rdm/metrics/api.py
@@ -8,7 +8,7 @@
 """ZenodoRDM Metrics API."""
 
 import calendar
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 
 import requests
 from flask import current_app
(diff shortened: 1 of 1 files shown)


site/zenodo_rdm/metrics/api.py
```diff
@@ -87,26 +87,27 @@ def get_uptime():
...
-
-        res = requests.post(
-            url,
-            json={
-                "api_key": api_key,
-                "custom_uptime_ranges": f"{start_ts}_{end_ts}",
-            },
-        )
+        if api_key:
+            end = datetime.now(timezone.utc).replace(
```
ptamarit, line 91: I would put this change from `utcnow` to `now` as a separated commit, since it's unrelated and the indentation change can make it difficult to notice depending on the diffing tool used.

ptamarit (approved):

PR #830, opened 2024-04-12
bug: check legacy_license is defined

carlinmack:
255 associated events, ZENODO-RDM-1WN

### DIFF
--- a/site/zenodo_rdm/legacy/deserializers/metadata.py
+++ b/site/zenodo_rdm/legacy/deserializers/metadata.py
@@ -257,8 +257,9 @@ class MetadataSchema(Schema):
         else:
             # If license does not exist in RDM, it is added as custom
             legacy_license = LEGACY_LICENSES.get(obj)
-            if legacy_license:
-                ret = {"title": {"en": legacy_license["title"]}}
+            if not legacy_license:
+                raise ValidationError(f"Invalid license provided: {obj}")
+            ret = {"title": {"en": legacy_license["title"]}}
 
         return [ret]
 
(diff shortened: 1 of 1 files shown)


site/zenodo_rdm/legacy/deserializers/metadata.py
```diff
@@ -257,7 +257,8 @@ def load_rights(self, obj):
         else:
             # If license does not exist in RDM, it is added as custom
             legacy_license = LEGACY_LICENSES.get(obj)
-            ret = {"title": {"en": legacy_license["title"]}}
+            if legacy_license:
+                ret = {"title": {"en": legacy_license["title"]}}
```
alejandromumo, line 261: It is not so bad that it blows up here, since this is the last chance to set the license, that is required to publish the record in RDM. Could we raise a Marshmallow exception? Looks better than python blowing up. e.g:

```python
if not legacy_license:
    raise ValidationError() 
```

For context, this is how we try to figure out the license:

1) Check if the license is an [alias](https://github.com/zenodo/zenodo-rdm/blob/master/legacy/zenodo_legacy/licenses.py#L19)
2) Try to [map the license](https://github.com/zenodo/zenodo-rdm/blob/35616a18b059c9196d8fe3966d0930e9f5ea242d/legacy/zenodo_legacy/data/legacy_to_rdm_map.json#L4) to one in RDM
3) If not defined yet, fallback to a [legacy license](https://github.com/zenodo/zenodo-rdm/blob/35616a18b059c9196d8fe3966d0930e9f5ea242d/legacy/zenodo_legacy/data/legacy_licenses.json#L4) but added it as a custom one in RDM



This issue appears on Github because users declare their licenses wrongly. E.g. `GPLv3+` is not a valid identifier (check [spdx](https://spdx.org/licenses/)) nor an alias (check legacy Zenodo codebase)

alejandromumo (commented):
LGTM but also requested the review of @slint , since he also mentioned that the license could be added as a custom during a standup. 

Also: you need to fix black formatting errors, CI is failing due to it.

slint (approved):
LGTM, at least so that users know what went wrong. Regarding being more flexible that's something we should look into later.

PR #832, opened 2024-04-15
fix: csv: change delimiter to comma to pass CSV dialect sniffing

ptamarit:
Running `invenio-cli services setup --force` fails on line 38 (a line with double quotes) in the new CSV dialect sniffing logic added in https://github.com/inveniosoftware/invenio-rdm-records/pull/1721.

Also, going to https://github.com/zenodo/zenodo-rdm/blob/master/app_data/vocabularies/licenses.csv, GitHub shows:
> We can make this file beautiful and searchable if this error is corrected: Illegal quoting in line 38.

Instead of just fixing the problem on this line, I converted the file to use comma delimiters as done in https://github.com/inveniosoftware/invenio-rdm-records/pull/1721.

I used the following snippet:

```python
import csv

with open(r"licenses.csv") as in_file, open(r"licenses-out.csv", 'w') as out_file:
    semicolonin = csv.reader(in_file, delimiter=';')
    commaout = csv.writer(out_file, delimiter=',')
    for row in semicolonin:
        commaout.writerow(row)
```

This seems to fix the fixture loading problem and the GitHub preview.

Remark: we are not including the URL changes from https://github.com/inveniosoftware/invenio-rdm-records/pull/1721 in `zenodo-rdm`.

### DIFF
--- a/app_data/vocabularies/licenses.csv
+++ b/app_data/vocabularies/licenses.csv
@@ -1,446 +1,446 @@
-id;title__en;description__en;icon;tags;props__url;props__scheme;props__osi_approved
-0bsd;BSD Zero Clause License;;;all,software;http://landley.net/toybox/license.html;spdx;y
-aal;Attribution Assurance License;;;all,software;https://opensource.org/licenses/attribution;spdx;y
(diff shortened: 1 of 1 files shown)


alejandromumo (approved):
Tested locally, the new `csv.Sniffer` code works

PR #833, opened 2024-04-16
docker: added iipserver service

alejandromumo:
* added iipserver service to docker-compose.yml
* enabled IIPServerProxy in invenio.cfg

closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1728

### DIFF
--- a/docker-services.yml
+++ b/docker-services.yml
@@ -107,11 +107,11 @@ services:
       - "DISABLE_SECURITY_DASHBOARDS_PLUGIN=true"
   iipserver:
     image: registry.cern.ch/docker.io/iipsrv/iipsrv:latest
-    ports: 
+    ports:
       - "8080:80"
     expose:
       - "8080"
     environment:
       - "CORS=*"
     volumes:
-      - ${INSTANCE_PATH}/iiif/images:/images # INSTANCE_PATH must be defined in the environment (or .env file)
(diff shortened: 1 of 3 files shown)


docker-services.yml
```diff
@@ -105,3 +105,13 @@ services:
...
+  iipserver:
+    image: registry.cern.ch/docker.io/iipsrv/iipsrv:latest
+    ports: 
+      - "8080:80"
+    expose:
+      - "8080"
+    environment:
+      - "CORS=*"
+    volumes:
+      - ${INSTANCE_PATH}/iiif/images:/images # INSTANCE_PATH must be defined in the environment (or .env file)
```
alejandromumo, line 117: This is a big assumption, currently, the instance path is stored in `./.invenio.private` (not a `.env` file). Any suggestions on how to enforce this?

docker-services.yml
```diff
@@ -105,3 +105,13 @@ services:
     environment:
       - "OPENSEARCH_HOSTS=http://search:9200"
       - "DISABLE_SECURITY_DASHBOARDS_PLUGIN=true"
+  iipserver:
+    image: registry.cern.ch/docker.io/iipsrv/iipsrv:latest
```
ptamarit, line 109: Wondering if we should pin to version 1.2.

ptamarit (approved):

docker-services.yml
slint: For the OpenShift image, we'll be pinning to specific versions, but for local dev it's a good chance to possibly catch breaking changes.

Issue #819, opened 2024-04-08
Finalize Mirador 3 JS distribution


slint:
```[tasklist]
### Tasks
- [ ] Coordinate with Dan (Data Futures) how the final distribution of the Mirador 3 previewer will look like
- [x] Make Mirador 3 previewer initialization configurable
```
Mirador-dist PR: https://github.com/data-futures/mirador3-dist-builder/pull/3

0einstein0:
I'll add the PR for mirador distribution if needed, after dicussion with Dan

PR #836, opened 2024-04-19
previewer: added mirador previewer

0einstein0:
Closes Issue https://github.com/zenodo/zenodo-rdm/issues/819
Partially Closes https://github.com/zenodo/zenodo-rdm/issues/840

Depends on: https://github.com/inveniosoftware/invenio-app-rdm/pull/2652
Depends on: https://github.com/inveniosoftware/invenio-app-rdm/pull/2657 

Screenshot 2024-04-19 at 15 18 00

Screenshot 2024-05-07 at 13 30 05
Screenshot 2024-05-07 at 13 30 12

### DIFF
--- a/assets/less/zenodo-rdm/globals/site.overrides
+++ b/assets/less/zenodo-rdm/globals/site.overrides
@@ -164,3 +164,8 @@ a.inverted {
 .cookie-banner.hidden {
   display: none !important;
 }
+
+.preview-iframe {
+  height:60vh;
+}
+
--- a/assets/less/zenodo-rdm/globals/site.overrides
+++ b/assets/less/zenodo-rdm/globals/site.overrides
@@ -166,12 +166,6 @@ a.inverted {
 }
 
 .preview-iframe {
-  height:600px;
+  height:60vh;
 }
 
-#previewer-banner{
-  display: none;
-    .message {
-    padding: 1em;
-  }
-}
--- a/assets/less/zenodo-rdm/globals/site.overrides
+++ b/assets/less/zenodo-rdm/globals/site.overrides
@@ -166,6 +166,12 @@ a.inverted {
 }
 
 .preview-iframe {
-  height:60vh;
+  height:600px;
 }
 
+#previewer-banner{
+  display: none;
+    .message {
+    padding: 1em;
+  }
+}
--- a/invenio.cfg
+++ b/invenio.cfg
@@ -825,13 +825,13 @@ TILES_STORAGE_PATH = "images/"  # relative to the instance path
 # See https://github.com/inveniosoftware/invenio-previewer/blob/master/invenio_previewer/config.py  # noqa
 PREVIEWER_PREFERENCE = [
     "csv_papaparsejs",
+    "pdfjs",
     "mirador",
     "iiif_simple",
     "simple_image",
     "json_prismjs",
     "xml_prismjs",
     "mistune",
-    "pdfjs",
     "video_videojs",
     "audio_videojs",
     "ipynb",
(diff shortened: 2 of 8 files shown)


invenio.cfg
ptamarit, line 1: `MIRADOR_PREVIEW_EXTENSIONS` should have `gif` removed, and `tiff` added.

assets/less/zenodo-rdm/globals/site.overrides
```diff
@@ -164,3 +164,14 @@ a.inverted {
 .cookie-banner.hidden {
   display: none !important;
 }
+
+.preview-iframe {
+  height:60vh;
```
ptamarit, line 169: Maybe to be tested on a few different viewports with a few different file types (e.g. zip) and validate that the UX is fine.

site/zenodo_rdm/previewer/mirador.py
```diff
@@ -19,8 +19,14 @@ def can_preview(file):
...
     supported_extensions = ["." + ext for ext in preview_extensions]
+    media_file_name = file.filename + ".ptif"
+
     return (
-        file.has_extensions(*supported_extensions) and file.record.data["is_published"]
+        file.has_extensions(*supported_extensions)
+        and not file.record.data["is_draft"]
+        and file.record._record.media_files.enabled
+        and file.record._record.media_files[media_file_name].processor["status"]
+        == "finished"
```
ptamarit, line 29: To be checked if records with private files are returning True or not (they should not).

ptamarit (approved):

assets/less/zenodo-rdm/globals/site.overrides
ptamarit, line 169: @carlinmack suggest to add a `min-height: 400px;` for small viewports.

invenio.cfg
```diff
@@ -841,20 +841,55 @@ PREVIEWER_PREFERENCE = [
...
     "workspace": {
         "draggingEnabled": True,
         "showZoomControls": True,
     }
 }
 """Config for the Mirador previewer."""
+
+IIIF_TILES_GENERATION_ENABLED = True
+
+IIIF_TILES_VALID_EXTENSIONS = [ "pdf","png", "jp2", "jpeg", "jpg", "png", "tif", "tiff"]
```
slint, line 892: comment: I think these get eventually normalized, but maybe it's worth being explicit about all extension types.

site/zenodo_rdm/previewer/mirador.py
```diff
@@ -19,15 +22,32 @@ def can_preview(file):
...
     """Render template."""
+    media_file_name = file.filename + ".ptif"
+
+    ext = splitext(file.filename)[1].lower()[1:]
+    format = (
+        ext
+        if ext in current_app.config["IIIF_SIMPLE_PREVIEWER_NATIVE_EXTENSIONS"]
+        else "jpg"
+    )
+    size = LocalProxy(lambda: current_app.config["IIIF_SIMPLE_PREVIEWER_SIZE"])
```
slint, line 39:
```suggestion
    size = current_app.config["IIIF_SIMPLE_PREVIEWER_SIZE"]
```

minor: since we're in the application context now, no need for a `LocalProxy`

site/zenodo_rdm/previewer/mirador.py
```diff
@@ -19,15 +22,32 @@ def can_preview(file):
...
+    url = f"{file.data['links']['iiif_base']}/full/{size}/0/default.{format}"
```
slint, line 40:
```suggestion
    iiif_simple_url = f"{file.data['links']['iiif_base']}/full/{size}/0/default.{format}"
```

nit: so that we know that it's for the "fallback" simple IIIF use

site/zenodo_rdm/previewer/mirador.py
```diff
@@ -19,15 +22,32 @@ def can_preview(file):
...
     return render_template(
         "invenio_app_rdm/records/mirador_preview.html",
+        css_bundles=["mirador-previewer.css"],
         file=file,
+        file_url=url,
+        media_file=(
+            file.record._record.media_files[media_file_name]
+            if file.record._record.media_files.enabled
+            else None
+        ),
```
slint, line 51:
```suggestion
        media_file=(
            file.record._record.media_files[media_file_name]
            if file.record._record.media_files.enabled
            else None
        ),
```

- This might fail if the file is not there at all, in which case maybe we add either a `media_file_name in file.record._record.media_files` check or use `file.record._record.media_files.get(...)`
- I would extract this in a variable

slint (commented):
LGTM! Some minor things for clarity

site/zenodo_rdm/previewer/mirador.py
0einstein0: If we use the file.record._record.media_files.get(...) we get `{"status": 403, "message": "Files are not enabled."}` error since we're trying to access media_files when they have not been enabled

invenio.cfg
0einstein0: would `IIIF_TILES_VALID_EXTENSIONS =["pdf","png", "jpeg", "jpg", "tiff"]` suffice?

site/zenodo_rdm/previewer/mirador.py
slint: I would keep the `file.record._record.media_files.enabled` check, but also be more "defensive" in case the file is not there (e.g. for existing Zenodo records that haven't been scheduled for tiles generation).

invenio.cfg
slint: I somewhat like having them "expanded". E.g. I'm not sure if we include `jp2` in the original config, and now it's too late to fix :)

PR #844, opened 2024-04-23
iiif: generate pyramidal tifs on publish

yashlamba:
WIP, not the final implementation.


yashlamba:
Closed by https://github.com/inveniosoftware/invenio-rdm-records/pull/1735

Issue #852, opened 2024-05-10
edit files: allow users to modify files for non-zenodo DOIs


alejandromumo:
We want to allow a set of users to edit files for Non-Zenodo DOIs. 

Discussed with @slint and agreed on implementing something similar to the media files edition, i.e. we create an action to manage files and then create a role associated with the action. Lastly, we add this role to specific users so they can modify the files. 

When this is deployed, we still need to create the role in production and associate the users:

```terminal
$ invenio roles create files-manager
$ invenio access allow zenodo-manage-files role files-manager
$ invenio roles add <user_email> files-manager
```

PR #853, opened 2024-05-10
files: add file management action and permission

alejandromumo:
closes https://github.com/zenodo/zenodo-rdm/issues/852

* Add a new action and permission for file management.
* Fixed signature of lock_edit_record_published_files to accept draft as well as record.


site/zenodo_rdm/config.py
```diff
@@ -335,20 +335,18 @@
 REDIRECTOR_RULES.update(EXPORT_REDIRECTS)
 
 
-def lock_edit_record_published_files(service, identity, record=None):
+def lock_edit_record_published_files(service, identity, record=None, draft=None):
     """Custom conditions for file bucket lock."""
+    # Check if draft already has pids or not, otherwise pass the record
+    use_draft = draft and draft.pids is not None
```
slint, line 341: Can't we always pass the record? Since this permission is mainly checked when we want to edit an already published record

site/zenodo_rdm/generators.py
```diff
@@ -18,6 +18,9 @@
 media_files_management_action = action_factory("manage-media-files")
 media_files_permission = Permission(media_files_management_action)
 
+files_management_action = action_factory("zenodo-manage-files")
```
slint, line 21: minor: not a big fan of the name, since it hints to a more general "file manager" role. Maybe something like `manage-external-doi-files`?

slint (approved):
LGTM, minor comments on naming and if to pass record/draft to the permission check

site/zenodo_rdm/generators.py
alejandromumo: I agree it is a bit generic now. But in this case, the action itself (and the `Generator`) are not aware of DOIs at all. That comes from the conditional generator `IfExternalDOIRecord`

I can still change it though, but maybe drop a comment and say "To be used for ExternalDOIs only"?

Issue #876, opened 2024-06-13
Displaying annotations in mirador using wadm files


0einstein0:
- Make mirador display annotations based on the annotations file provided either in files or media_files
- Annotations are only readonly, no creation, edition or deletion

Example record on Zenodo: https://zenodo.org/records/10817390
Example Previewer: https://avisblatt.dg-basel.hasdai.org/records/bsf3z-wqy40

PR #879, opened 2024-06-19
mirador: added annotations support for previewer

0einstein0:
Closes Issue https://github.com/zenodo/zenodo-rdm/issues/876

### InvenioRDMAdapter
```
/** */
export default class InvenioRDMAdapter {
  /** */
  constructor(canvasId, annotations) {
    this.canvasId = canvasId;
    this.annotations = annotations
    this.readOnly = true;
  }

  /** */
  get annotationPageId() {

    // assume for now that canvas.jpg has an associated canvas.jpg.short.wadm 
    // ideally have an endpoint in /api/iiif which is more flexible

    let filename = this.canvasId.split('/').pop()
    let url = this.annotations[filename];
    return url;
  }

  /** */
  async all() {
    return (await fetch(this.annotationPageId)).json();
  }
}

```
## Annotation in Files
Screenshot 2024-06-19 at 15 45 04

## Annotations in Media Files
Screenshot 2024-06-19 at 13 44 09

## No Annotations
Screenshot 2024-06-19 at 13 50 52


invenio.cfg
```diff
@@ -868,36 +868,11 @@ MIRADOR_PREVIEW_EXTENSIONS = [
...
-            "layers": False,
-        },
-        "views": [
-            {"key": "single", "behaviors": ["individuals"]},
-            {"key": "book", "behaviors": ["paged"]},
-            {"key": "scroll", "behaviors": ["continuous"]},
-            {"key": "gallery"},
-        ],
-    },
+
```
0einstein0, line 871: since I am updating the flags in the `window` var in the mirador dist in default config based on existence of annotations, I removed it from here, as it overrides default config. If we'd rather have it here, then I can instead update the config in `mirador.py` instead

PR #893, opened 2024-06-25
subcommunities: added  request type in  Zenodo

### LINKED ISSUE #891: Modify backend to accept a project ID in the community inclusion request
Subcommunity requests created against the "EU" community require the project ID

alejandromumo:
closes #891

### DIFF
--- a/site/zenodo_rdm/subcommunities/schema.py
+++ b/site/zenodo_rdm/subcommunities/schema.py
@@ -1,0 +1,31 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2024 CERN.
+#
+# Zenodo-RDM is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+"""Subcommunities schema implementation."""
+
+from invenio_communities.subcommunities.services.schema import (
+    MinimalCommunitySchema,
+    SubcommunityRequestSchema,
+)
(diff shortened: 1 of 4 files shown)


site/zenodo_rdm/subcommunities/schema.py
```diff
@@ -0,0 +1,60 @@
...
+    project = fields.String()
+
+    @post_load(pass_original=True)
+    def process_project(self, data, original, **kwargs):
+        """Validate the project field.
+
+        For 'eu' community, it is a required field.
+        """
+        project = original.get("project")
+        is_ec_community = self.context.get("community_slug") == "eu"
```
alejandromumo, line 32: I am not convinced by this at all ... 

Having to hard-code the EU community slug here feels wrong, but we don't have a way to customize this request for specific communities yet. 

Perhaps it could "look better" if we add this as a config?

site/zenodo_rdm/subcommunities/schema.py
```diff
@@ -0,0 +1,60 @@
...
+        if is_ec_community and not project:
+            raise ValidationError(_("Project is required for EC communities."))
+        if project:
+            p_str = project.split("::", 1)
```
alejandromumo, line 36: I don't like this, however the `funder` field has to be explicitly set when creating the community. Otherwise, the community is created just with the award info (and no funder reference...)

PR #954, opened 2024-07-29
app_data: add funders, affiliations, awards datastream config

slint:


app_data/vocabularies-future.yaml
```diff
@@ -4,4 +4,73 @@ names:
...
+    - type: json
+  transformers:
+    - type: ror-funders
+  writers:
+    - args:
+        writer:
+          type: funders-service
+      type: async
+
+awards-fill:
```
ptamarit, line 39:
```suggestion
awards-full:
```

slint: There's a small caveat for this file, where basically the `--vocabulary <VOCAB>` parameter passed to `invenio vocabularies import ...` has to match the top-level key inside the file, but also has to be one of the keys defined [in the datastream factory dictionary](https://github.com/inveniosoftware/invenio-vocabularies/blob/cf8243c161c628ae03a5711a1320976508142406/invenio_vocabularies/factories.py#L101-L105). I.e. the following wouldn't work:

```bash
invenio vocabularies --file app_data/vocabularies-future.yaml --vocabulary awards-full
```

IMHO, it's a limitation and wrong usage of the CLI, but in any case since we're moving to Jobs it doesn't matter that much now. IIRC we also discussed about adding a new option entirely to the CLI to tackle this issue

ptamarit: In this case, since both configs only differ regarding `origin` which can be set via the `cli`, I think the following would achieve the desired result:

```
invenio vocabularies --file app_data/vocabularies-future.yaml --vocabulary awards -o full
```

slint: Ah, nice, I hadn't thought of that 😅 

Then I think we're good to just keep the one entry; I'll remove the other one and keep the `full` parameter as a default, so that for the incremental one just has to call `--origin diff`

PR #962, opened 2024-08-08
installation: bump dependencies

0einstein0:


slint:
Bumped and pushed separately, since we were missing a release on `invenio-records-resources`

PR #965, opened 2024-08-15
Various small fixes

slint:
- **ci: update workflow config and split JS from Python tests**
- **config: set 50GB bucket quota via old config**
- **records-ui: update link to citations FAQ**

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,55 +25,57 @@ on:
         default: 'Manual trigger'
 
 jobs:
-  Tests:
-    runs-on: ubuntu-20.04
-    continue-on-error: ${{ matrix.ignore-failure }}
+  Python:
+    runs-on: ubuntu-24.04
     strategy:
       matrix:
           package-dir: ["site", "legacy"]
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -25,49 +25,34 @@ on:
...
-        run: |
-          pip install "pipenv==2023.12.1"
+          cache: pip
+          cache-dependency-path: Pipfile.lock
 
       - name: Install dependencies
         run: |
           sudo apt-get update
           sudo apt-get -y install libkrb5-dev libvips-dev
+          pip install "pipenv==2023.12.1"
```
ptamarit, line 55: Should we document why we are pining the `pipenv` version here?

ptamarit (approved):

.github/workflows/tests.yml
slint: Yes, in the end actually this wasn't necessary, but we wanted to match the `pipenv` version that's on the Docker image that is deployed.

I'll add a comment

ptamarit (approved):

PR #972, opened 2024-08-22
mirador: updating iiif fallback preview conditions

0einstein0:
- Updating file preview link if the file size is less than 256x256, or metadata is not present
_file_url: ..../records/230/files/200x210.png?preview=0_
Screenshot 2024-08-22 at 17 25 12

###
- _If tile status is processing for more than an hour and it's size is bigger than configured:_
Screenshot 2024-08-29 at 15 37 27

###
- _If tile status is processing for more than an hour and it's size is less than configured:_
  - _file_url: .../api/iiif/record:230:200x210.png/full/full/0/default.png_
Screenshot 2024-08-23 at 10 15 57


site/zenodo_rdm/previewer/mirador.py
```diff
@@ -64,6 +55,46 @@ def preview(file):
         if tile_file is not None and tile_file.processor:
             tile_status = tile_file.processor.get("status")
 
+    # Check if the file was updated more than an hour ago
+    last_updated_time = datetime.fromisoformat(file.data["updated"])
+    current_time = datetime.now(timezone.utc)
+    time_diff = current_time - last_updated_time
+
+    # If the image tile status size is processing and the image was last updated more than an hour ago
+    if tile_status == "processing" and time_diff > timedelta(hours=1):
```
anikachurilova, line 64: Should `timedelta(hours=1)` be placed in the config?

site/zenodo_rdm/previewer/mirador.py
```diff
@@ -64,6 +55,46 @@ def preview(file):
...
+            return render_template(
+                "invenio_app_rdm/records/previewers/simple_image_preview.html",
+                css_bundles=["mirador-previewer.css"],
+                file_url=file.file.links["iiif_api"],
+            )
+        # The image size is greater than configured size,
+        # image cannot be previewed
+        elif file.size > current_app.config["PREVIEWER_MAX_IMAGE_SIZE_BYTES"]:
+            return render_template(
+                "invenio_app_rdm/records/previewers/default.html",
```
ptamarit, line 76: I feel like the `default` is not a great name in this case. Maybe something like?

```suggestion
                "invenio_app_rdm/records/previewers/not_available_preview.html",
```

site/zenodo_rdm/assets/semantic-ui/less/zenodo_rdm/previewer/mirador-previewer.less
```diff
@@ -22,6 +23,12 @@
     box-shadow: 0px 0px 0px 1px #A9D5DE inset, 0px 0px 0px 0px rgba(0, 0, 0, 0);
 }
 
+.image-container {
```
ptamarit, line 26: Maybe using a name which suggest that it's not related to mirador?

```suggestion
.simple-image-container {
```

site/zenodo_rdm/assets/semantic-ui/less/zenodo_rdm/previewer/mirador-previewer.less
```diff
@@ -39,4 +46,8 @@ i.info.icon {
     content: "i";
     color: #F8FFFF; 
     font-weight: 600;
+  }
+
+  h3, p {
+    font-family: 'Lato','Helvetica', Arial, sans-serif;
```
ptamarit, line 52: I would scope these to the parent CSS class container so that the style does not leak.

ptamarit (approved):

site/zenodo_rdm/previewer/mirador.py
slint: Naming committee decided the config is:

```python
PREVIEWER_IMAGE_FAILED_PROCESSING_TIMEDELTA = {
    "hours": 1,
}
```

ptamarit (approved):

PR #974, opened 2024-08-30
config: enabled file-integrity checks

0einstein0:
Closes Issue: https://github.com/zenodo/ops/issues/402

### DIFF
--- a/invenio.cfg
+++ b/invenio.cfg
@@ -159,7 +159,6 @@ APP_DEFAULT_SECURE_HEADERS = {
 # ======
 
 # Remove file integrity checks
-CELERY_BEAT_SCHEDULE.pop("file-checks", None)
 CELERY_BEAT_SCHEDULE.pop("file-integrity-report", None)
 
 CELERY_BEAT_SCHEDULE = {
@@ -190,19 +189,6 @@ CELERY_BEAT_SCHEDULE = {
         "task": "invenio_swh.tasks.cleanup_depositions",
         "schedule": crontab(minute=0, hour=6), # Every day at 06:00 UTC
     },
-    'file-checks': {
-        'task': 'invenio_files_rest.tasks.schedule_checksum_verification',
-        'schedule': timedelta(hours=1),
-        'kwargs': {
-            'batch_interval': {'hours': 1},
-            'frequency': {'days': 14},
-            'max_count': 0,
-            # Query taking into account only files with URI prefixes defined by
-            # the FILES_REST_CHECKSUM_VERIFICATION_URI_PREFIXES config variable
-            'files_query':
-                'zenodo_rdm.files.checksum_verification_files_query',
-        },
-    },
     'hard-file-checks': {
         'task': 'invenio_files_rest.tasks.schedule_checksum_verification',
         'schedule': timedelta(hours=1),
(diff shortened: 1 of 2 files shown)


site/zenodo_rdm/files.py
```diff
@@ -149,3 +151,21 @@ def send_file(
...
     """Create custom storage factory to enable file offloading."""
     return pyfs_storage_factory(filestorage_class=EOSFilesOffload, **kwargs)
+
+
+def checksum_verification_files_query():
+    """Return a FileInstance query taking into account file URI prefixes."""
+    files = FileInstance.query
+
+    uri_prefixes = current_app.config.get(
+        "FILES_REST_CHECKSUM_VERIFICATION_URI_PREFIXES"
```
alejandromumo, line 161: Is this something we need to set in prod?

invenio.cfg
```diff
@@ -189,7 +189,37 @@ CELERY_BEAT_SCHEDULE = {
...
+            # Manually check and calculate checksums of files biannually
+            'frequency': {'days': 180},
+            # Split batches based on total files size
+            'max_size': 0,
+            # Query taking into account only files with URI prefixes defined by
+            # the FILES_REST_CHECKSUM_VERIFICATION_URI_PREFIXES config variable
+            'files_query':
+                'zenodo_rdm.files.checksum_verification_files_query',
+            # Actual checksum calculation, instead of relying on a EOS query
+            'checksum_kwargs': {'use_default_impl': True},
```
alejandromumo, line 220: I don't find where `use_default_impl` is implemented, can you share how that works?

slint: This is copy-pasted from the original [config we had in legacy](https://github.com/zenodo/zenodo/blob/482ee72ad501cbbd7f8ce8df9b393c130d1970f7/zenodo/config.py#L240-L241). The argument is used [in the XRootD storage class](https://github.com/inveniosoftware/invenio-xrootd/blob/5d1507fc85619958ce8bd0f038bfd116143f06fb/invenio_xrootd/storage.py#L64), to basically force it to "manually" compute the MD5 checksum on our side (instead of relying on EOS).

site/zenodo_rdm/files.py
```diff
@@ -149,3 +151,21 @@ def send_file(
...
+    )
+    if uri_prefixes:
+        files = files.filter(
+            sa.or_(*[FileInstance.uri.startswith(p) for p in uri_prefixes])
+        )
+    # Filtering out .ptif media files
+    excluded_path = current_app.config["IIIF_TILES_STORAGE_BASE_PATH"]
+    files = files.filter(~FileInstance.uri.contains(excluded_path))
```
slint, line 169:
```suggestion
    files = files.filter(~FileInstance.uri.startswith(excluded_path))
```

minor: I think this will make for a more efficient SQL query also, i.e. the difference between `LIKE '%/images/iiif/%` vs `LIKE `/images/iiif/%`.

site/zenodo_rdm/files.py
```diff
@@ -149,3 +151,21 @@ def send_file(
...
+    uri_prefixes = current_app.config.get(
+        "FILES_REST_CHECKSUM_VERIFICATION_URI_PREFIXES"
+    )
+    if uri_prefixes:
+        files = files.filter(
+            sa.or_(*[FileInstance.uri.startswith(p) for p in uri_prefixes])
+        )
+    # Filtering out .ptif media files
+    included_path = current_app.config["FILES_STORAGE_BASE_PATH"]
+    files = files.filter(FileInstance.uri.startswith(included_path))
```
slint, line 169: I had in mind using the `FILES_REST_CHECKSUM_VERIFICATION_URI_PREFIXES` config entirely, so no need to have another config variable.

alejandromumo:
We have parts of this code in app-rdm by the way:

https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/utils/files.py#L22

https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/config.py#L433

alejandromumo (approved):

PR #1029, opened 2024-10-18
 horizon: frontpage: disable autofocus on CommunitySelectionSearch & bring back projects to frontpage

ptamarit:
Partially closes inveniosoftware/invenio-communities#1227
Depends on inveniosoftware/invenio-rdm-records#1848

### Description

The "Search in all communities" input in the community front page is currently auto-focused (see linked issue).
While it makes sense for the auto-focus to be present in a model, it does not in a community front page where this is not the only input and it is lower in the page.


slint (approved):

ptamarit:
Just rebased (although it's not really necessary).

PR #1047, opened 2024-11-04
moderation: added rule for excessive header tags

0einstein0:
Closes Issue https://github.com/zenodo/zenodo-rdm/issues/1044

### DIFF
--- a/site/zenodo_rdm/moderation/rules.py
+++ b/site/zenodo_rdm/moderation/rules.py
@@ -74,9 +74,18 @@ def links_rule(identity, draft=None, record=None):
 
 
 def text_sanitization_rule(identity, draft=None, record=None):
-    """Calculate a score for excessive emoji usage in metadata text."""
+    """Calculate a score based on excessive emoji and HTML tag usage in metadata text."""
     record_text = " ".join(map(str, record.metadata.values()))
-    return current_scores.spam_emoji if len(extract_emojis(record_text)) > 3 else 0
+    htag_count = len(re.findall(r"<h[1-9]\b[^>]*>", record_text, re.IGNORECASE))
+    score = 0
+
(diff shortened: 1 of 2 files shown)


site/zenodo_rdm/moderation/rules.py
```diff
@@ -74,9 +74,18 @@ def links_rule(identity, draft=None, record=None):
...
 def text_sanitization_rule(identity, draft=None, record=None):
-    """Calculate a score for excessive emoji usage in metadata text."""
+    """Calculate a score based on excessive emoji and HTML tag usage in metadata text."""
     record_text = " ".join(map(str, record.metadata.values()))
-    return current_scores.spam_emoji if len(extract_emojis(record_text)) > 3 else 0
+    htag_count = len(re.findall(r"<h[1-9]\b[^>]*>", record_text, re.IGNORECASE))
+    score = 0
+
+    if len(extract_emojis(record_text)) > 3:
+        score += current_scores.spam_emoji
```
anikachurilova, line 83: I am not sure I understand why we add here the default value? Why do we add 5 if there are 3 emojis?

0einstein0: The scoring is a bit arbitrary and not associated with the number of emojis. It's just how much weightage we give to the occurrence of a specific element in the record.

kpsherva: maybe it makes sense to drop a comment explaining the number is arbitrary, otherwise the next person who develops something might also question it

slint: I think all "magic" numbers we have in the rules should be moved to the `MODERATION_SCORES` config dictionary. They will be tweaked in any case at some point until we find a good set of defaults.

Merging this one now, since we'll have another iteration soon to refine the values.

PR #1048, opened 2024-11-05
moderation: move domains ban/safe-list to DB table

### LINKED ISSUE #1043: Convert banned/safe links domains config to a DB table
We want to store all the link domains in a table, and mark their status (banned/safe). Ideally when matching, we should be querying the DB like so:

```python
        domain_parts = self.extract_domain(url)  # ['io', 'github', 'slint']

        """
        SELECT status
        FROM domain_tree
        WHERE domain LIKE 'io.github%'  -- this is a "good" domain and should match
        
        -- This is a "bad" more specific domain
        -- WHERE domain LIKE 'io.github.ketogummies'
        """
```

slint:
- Closes #1043


site/zenodo_rdm/moderation/rules.py
```diff
@@ -65,11 +64,19 @@ def links_rule(identity, draft=None, record=None):
...
+        if domain.status == LinkDomainStatus.BANNED:
+            if domain.score is not None:
+                score += domain.score
+            else:
+                score += current_scores.spam_link
+        elif domain == LinkDomainStatus.SAFE:
+            if domain.score is not None:
+                score += domain.score
+            else:
+                score += current_scores.ham_link
```
sakshamarora1, line 79: How do we score if the status is MODERATED?

sakshamarora1 (commented):
LGTU with @0einstein0
Just one question

site/zenodo_rdm/moderation/rules.py
slint: In practice, they contribute `0`, since it means that we have to have a look at them (i.e. "moderate" them). 

So tbh, I added the `MODERATED` status just to have a "other" value that can be used as a placeholder for ingesting new domains that we don't want to mark as banned/safe immediately.

sakshamarora1 (approved):

PR #1049, opened 2024-11-06
moderation: added query match rule

0einstein0:
Closes Issue https://github.com/zenodo/zenodo-rdm/issues/1042


site/zenodo_rdm/moderation/config.py
```diff
@@ -46,12 +52,14 @@
...
     text_sanitization_rule,
+    match_query_rule,
 ]
 """Scoring rules for record moderation."""
 
 MODERATION_COMMUNITY_SCORE_RULES = [
     links_rule,
     text_sanitization_rule,
     verified_user_rule,
+    match_query_rule,
```
slint, line 63: minor: I think for communities we'll need a different rule/config, since the mappings are different

site/zenodo_rdm/moderation/rules.py
```diff
@@ -123,3 +124,44 @@ def files_rule(identity, draft=None, record=None):
...
+    )
+
+
+def community_match_query_rule(identity, draft=None, record=None):
+    """Match query rule for communities."""
+    return match_query_rule(
+        identity,
+        draft=draft,
+        record=record,
+        index=current_app.config.get("MODERATION_COMMUNITY_PERCOLATOR_INDEX"),
```
slint, line 166:
```suggestion
        index=current_moderation.record_queries_index,
```

minor: and then we can calculate it with something like:

```python
from invenio_search.utils import build_alias_name

record_queries_index = build_alias_name(f"moderation-queries-{record_cls.index._name}")
```

site/zenodo_rdm/moderation/percolator.py
```diff
@@ -0,0 +1,84 @@
...
+
+"""Percolator."""
+
+
+from flask import current_app
+from invenio_search import current_search_client
+from invenio_search.utils import build_alias_name, build_index_name
+
+
+def create_percolator_index(index_name, record_cls):
```
slint, line 18:
```suggestion
def create_percolator_index(record_cls):
```

minor: the percolator index prefix should be in config (e.g. `MODERATION_PERCOLATOR_INDEX_PREFIX = "moderation-queries-"`). We can then reuse the config below to avoid hard-coding.

For the `index_name` param, we can either remove it entirely or make it optional (and default to the config value)

site/zenodo_rdm/moderation/percolator.py
```diff
@@ -0,0 +1,84 @@
...
+
+    record_index = build_alias_name(record_cls.index._name)
+    combined_index_name = f"{index_name}-{record_cls.index._name}"
+    # Build the name for the new percolator index, using a prefix and the record's index name
+    percolator_index = build_index_name(combined_index_name, app=current_app)
+
+    # Get the current mapping for the record index to copy its structure
+    record_mapping = current_search_client.indices.get_mapping(index=record_index)
+    # Extract the mappings from the record index and store in `percolator_mappings`
+    percolator_mappings = list(record_mapping.values())[0]["mappings"]
```
slint, line 35:
```suggestion
    assert len(record_mapping) == 1
    percolator_mappings = list(record_mapping.values())[0]["mappings"]
```

minor: just to make sure we're picking the right mapping

site/zenodo_rdm/moderation/percolator.py
```diff
@@ -0,0 +1,84 @@
...
+            )
+        except Exception as e:
+            current_app.logger.exception(e)
+
+
+def index_percolate_query(record_cls, query_string, active=True, score=1.0, notes=None):
+    """Index a percolate query."""
+    try:
+        current_search_client.index(
+            index=build_alias_name(f"moderation-queries-{record_cls.index._name}"),
```
slint, line 75: minor: since this seems to be called often, I would extract to a util function

site/zenodo_rdm/moderation/rules.py
```diff
@@ -123,3 +125,22 @@ def files_rule(identity, draft=None, record=None):
...
+
+
+def match_query_rule(identity, draft=None, record=None):
+    """Calculate a score based on matched percolate queries against the given document in the specified index."""
+
+    document = record.dumps()
+    percolator_index = build_alias_name(f"moderation-queries-{record.index._name}")
+    matched_queries = current_search_client.search(
+        index=percolator_index,
+        body={"query": {"percolate": {"field": "query", "document": document}}},
```
slint, line 137: minor: also add a filter so that we only match `active: true` percolator queries

site/tests/moderation/test_moderation_queries.py
```diff
@@ -0,0 +1,36 @@
...
+
+"""Test ModerationQuery model class."""
+
+
+from invenio_db import db
+from invenio_search import current_search_client
+from zenodo_rdm.moderation.models import ModerationQuery
+
+
+def test_moderation_query_creation(app):
```
slint, line 16: minor/shelve: I trust the feature works, but if relatively easy I would also add a test that performs a match query.

slint (commented):
Some minor things, but otherwise LGTM

site/zenodo_rdm/moderation/models.py
```diff
@@ -73,3 +79,49 @@ def lookup_domain(cls, url):
...
+
+    @classmethod
+    def create(
+        cls, query_string, record_cls=ZenodoRDMRecord, notes=None, score=0, active=True
+    ):
+        """Create a new moderation query with a configurable record class."""
+        query = cls(query_string=query_string, notes=notes, score=score, active=active)
+        db.session.add(query)
+
+        index_percolate_query(record_cls, query_string, active, score, notes)
```
slint, line 112: Move out, and add a new CLI command to add a single query (so we have the usage)

slint (approved):

site/zenodo_rdm/cli.py
```diff
@@ -246,3 +258,115 @@ def delete_record(recid):
...
+
+
+def create_and_index_query(record_cls, query_string, notes, score, active):
+    """Create and index a single moderation query."""
+    query = ModerationQuery.create(
+        query_string=query_string, notes=notes, score=score, active=active
+    )
+
+    db.session.commit()
+    index_percolate_query(record_cls, query.id, query_string, active, score, notes)
```
slint, line 372:
```suggestion
    index_percolate_query(record_cls, query)
```

minor/shelve: since all the fields are in the `query` object

PR #1057, opened 2024-11-11
moderation: add rule results to dict and eval

yashlamba:
Refactor to accommodate ML model results and better logging.

### DIFF
--- a/site/zenodo_rdm/moderation/config.py
+++ b/site/zenodo_rdm/moderation/config.py
@@ -45,21 +45,21 @@ MODERATION_SPAM_FILE_EXTS = {"jpg", "jpeg", "pdf", "png", "jfif", "docx", "webp"
 MODERATION_EXEMPT_USERS = []
 """List of users exempt from moderation."""
 
-MODERATION_RECORD_SCORE_RULES = [
-    verified_user_rule,
-    links_rule,
-    files_rule,
-    text_sanitization_rule,
-    match_query_rule,
-]
+MODERATION_RECORD_SCORE_RULES = {
+    "verified_user_rule": verified_user_rule,
+    "links_rule": links_rule,
+    "files_rule": files_rule,
+    "text_sanitization_rule": text_sanitization_rule,
+    "match_query_rule": match_query_rule,
+}
 """Scoring rules for record moderation."""
 
-MODERATION_COMMUNITY_SCORE_RULES = [
-    links_rule,
-    text_sanitization_rule,
-    verified_user_rule,
-    match_query_rule,
-]
+MODERATION_COMMUNITY_SCORE_RULES = {
+    "links_rule": links_rule,
+    "text_sanitization_rule": text_sanitization_rule,
+    "verified_user_rule": verified_user_rule,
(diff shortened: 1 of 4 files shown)


site/zenodo_rdm/moderation/config.py
```diff
@@ -41,21 +41,21 @@
...
-    text_sanitization_rule,
-    verified_user_rule,
-    match_query_rule,
-]
+MODERATION_COMMUNITY_SCORE_RULES = {
+    "links-rule": links_rule,
+    "text-sanitization-rule": text_sanitization_rule,
+    "verified-user-rule": verified_user_rule,
+    "match-query-rule": match_query_rule,
+}
```
yashlamba, line 58: We can also use class based rules to avoid this and maintaining a sequence in rules.

site/zenodo_rdm/moderation/handlers.py
```diff
@@ -73,14 +73,17 @@ def rules(self):
         """Get scoring rules."""
         if isinstance(self._rules, str):
             return current_app.config[self._rules]
-        return self._rules or []
+        return self._rules or {}
+
+    def evaluate_result(self, params):
+        return sum(params.values())
```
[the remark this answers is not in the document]
slint: We'll change the `evaluate_result` once we have the spam rule in, so that it only sums the values that make sense as part of the score, and uses the spam probability separately "decision-tree-style".

site/zenodo_rdm/moderation/config.py
yashlamba: I think it's an easy refactor if we want to achieve that sometime, right now we don't have a sequence based use case.

site/zenodo_rdm/moderation/handlers.py
```diff
@@ -77,17 +81,19 @@ def run(self, identity, draft=None, record=None, user=None, uow=None):
...
-                score += rule(identity, draft=draft, record=record)
+            results = {}
+            for name, rule in self.rules.items():
+                results[name] = rule(identity, draft=draft, record=record)
 
             action_ctx = {
                 "user_id": user.id,
                 "record_pid": record.pid.pid_value,
-                "score": score,
+                "results": results,
```
slint, line 91: minor: there's a part in the `update_moderation_request` task that formats the score/results and includes it in a comment (the cheap way to log results basically)

site/zenodo_rdm/moderation/handlers.py
```diff
@@ -77,17 +81,20 @@ def run(self, identity, draft=None, record=None, user=None, uow=None):
...
+            for name, rule in self.rules.items():
+                results[name] = rule(identity, draft=draft, record=record)
 
+            evaluation = self.evaluate_result(results)
             action_ctx = {
                 "user_id": user.id,
                 "record_pid": record.pid.pid_value,
-                "score": score,
+                "results": results,
+                "final_evaluation": evaluation,
```
slint, line 93:
```suggestion
                "evaluation": evaluation,
```

nit: you never know when the "final" evaluation comes :)

PR #1058, opened 2024-11-11
Moderation fixes

### LINKED ISSUE #1046: Sync current rules with manual moderation scripts

slint:
- Closes #1046.
- **moderation: update workflow for exempt and verified users**
- **moderation: fix percolators index naming and CLI usage**

### DIFF
--- a/site/zenodo_rdm/moderation/handlers.py
+++ b/site/zenodo_rdm/moderation/handlers.py
@@ -71,6 +71,12 @@ class BaseScoreHandler:
     def run(self, identity, draft=None, record=None, user=None, uow=None):
         """Calculate the moderation score for a given record or draft."""
         try:
+            if user.id in self.exempt_users:
+                current_app.logger.debug(
+                    "User is exempt from moderation", extra={"user_id": user.id}
+                )
+                return
+
             score = 0
             for rule in self.rules:
                 score += rule(identity, draft=draft, record=record)
@@ -122,6 +128,10 @@ class BaseScoreHandler:
                         "Manual moderation action triggered",
                         extra=action_ctx,
                     )
+        # Re-raise UserBlockedException to prevent further processing of the record
+        except UserBlockedException:
+            raise
+        # Failsafe in the moderation check is faulty
         except Exception:
             current_app.logger.exception("Error calculating moderation score")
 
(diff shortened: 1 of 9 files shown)


site/zenodo_rdm/moderation/handlers.py
```diff
@@ -75,106 +58,136 @@ def rules(self):
...
+                    if user.verified:
+                        action_ctx["action"] = "moderate"
+                        self._moderate(user, uow, action_ctx)
+                    else:
+                        self._block(user, uow, action_ctx)
                 else:
                     current_app.logger.error(
                         "Block moderation action triggered",
                         extra=action_ctx,
                     )
```
yashlamba, line 106: this might be good to have when we are applying the action as well? in some other way maybe

site/zenodo_rdm/moderation/handlers.py
```diff
@@ -185,7 +198,30 @@ def __init__(self):
...
+        user_id = None
+        if identity == system_identity:
+            user_id = record.parent.access.owned_by.owner_id
+        else:
+            user_id = identity.id
+        if user_id is None:
+            current_app.logger.error(
+                "No user found for moderation action", stack_info=True
+            )
+            return
```
yashlamba, line 210: Should this be return or raise?

I can't think of when are we expecting this.

yashlamba (approved):
Few minor comments, rest looks good.

site/zenodo_rdm/moderation/handlers.py
slint: It's logged on the user moderation request as a comment for now (see the tasks file). But yeah, I would prefer later for it to be logged as structured application logging to OpenSearch

site/zenodo_rdm/moderation/handlers.py
slint: In general there's no "result" coming back form these calls, so the intent is to just "bail" early here.

slint: Also, not having a user is often when there's some complex workflow with nested service calls that somehow lose the `identity` on the way (which why above we use the record owner ID instead as a fallback)

site/zenodo_rdm/moderation/handlers.py
```diff
@@ -75,106 +58,136 @@ def rules(self):
...
-            user = UserAggregate.get_record(user_id)
             action_ctx = {
-                "user_id": user_id,
+                "user_id": user.id,
                 "record_pid": record.pid.pid_value,
                 "score": score,
             }
-            apply_actions = current_app.config.get("MODERATION_APPLY_ACTIONS", False)
+            current_app.logger.info(
+                f"Moderation score for {user.id} calculated: {score}", extra=action_ctx
```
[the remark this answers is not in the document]
slint: Actually, all the `info` calls should be `debug`... The issue I have is that `extra`s are not logged right when running the local dev server, which we should look into, since they can be very useful.

PR #1067, opened 2024-11-18
subcommunities: add invitation request

### LINKED ISSUE #1061: Create subcommunity invitation request
At current people can create a request from their community to the EU ORR to be added as a subcommunity

In the associated PRs, I add the ability to create an invitation from the EU ORR to a community to make it a subcommunity (the inverse action)

Done

- [x] Request is created
- [x] Emails are sent
- [x] Request can be accepted, declined or cancelled
- [x] Add screenshots to PRs
- [x] Double check what the emails look like
- [x] Expire action

To do
- [x] Fix sending to parent on creation bug
- [x] Fix TODO (passing name/award url)
- [ ] Add tests

Image

carlinmack:
Close #1061 

image

### DIFF
--- a/scripts/sleep.py
+++ b/scripts/sleep.py
@@ -1,0 +1,4 @@
+import time
+
+while True:
+    time.sleep(60 * 60)
--- a/site/zenodo_rdm/legacy/requests/utils.py
+++ b/site/zenodo_rdm/legacy/requests/utils.py
@@ -12,13 +12,13 @@ from datetime import datetime, timedelta
 from invenio_access.permissions import system_identity
 from invenio_records.dictutils import dict_lookup
 from invenio_records_resources.services.uow import unit_of_work
-from invenio_requests import current_request_type_registry, current_requests_service
+from invenio_requests import (current_request_type_registry,
+                              current_requests_service)
 from invenio_requests.resolvers.registry import ResolverRegistry
 from invenio_search.engine import dsl
 
 from .community_manage_record import CommunityManageRecord
 from .record_upgrade import LegacyRecordUpgrade
-from ...subcommunities.request import ZenodoSubCommunityInvitationRequest
 
 
 def exists(identity, record):
@@ -105,35 +105,6 @@ def submit_community_manage_record_request(user_id, uow, comment=None):
     )
 
     # submit the request
-    return current_requests_service.execute_action(
-        system_identity, request_item._request.id, "submit", data=comment, uow=uow
-    )
-
-
-@unit_of_work()
-def submit_join_as_subcommunity_request(
-    parent_community_uuid, community_uuid, uow, comment=None
-):
-    """Create and submit a SubCommunityInvitation request."""
-    type_ = current_request_type_registry.lookup(ZenodoSubCommunityInvitationRequest.type_id)
-    creator = ResolverRegistry.resolve_entity_proxy(
-        {"community": parent_community_uuid}
-    ).resolve()
-    receiver = ResolverRegistry.resolve_entity_proxy(
-        {"community": community_uuid}
-    ).resolve()
-    expires_at = datetime.utcnow() + timedelta(days=30)
-
-    request_item = current_requests_service.create(
-        system_identity,
-        {},
-        type_,
-        receiver,
-        creator=creator,
-        expires_at=expires_at,
-        uow=uow,
-    )
-
     return current_requests_service.execute_action(
         system_identity, request_item._request.id, "submit", data=comment, uow=uow
     )
--- a/site/zenodo_rdm/legacy/requests/utils.py
+++ b/site/zenodo_rdm/legacy/requests/utils.py
@@ -18,6 +18,7 @@ from invenio_search.engine import dsl
 
 from .community_manage_record import CommunityManageRecord
 from .record_upgrade import LegacyRecordUpgrade
+from ...subcommunities.request import ZenodoSubCommunityInvitationRequest
 
 
 def exists(identity, record):
@@ -104,6 +105,35 @@ def submit_community_manage_record_request(user_id, uow, comment=None):
     )
 
     # submit the request
+    return current_requests_service.execute_action(
+        system_identity, request_item._request.id, "submit", data=comment, uow=uow
+    )
+
+
+@unit_of_work()
+def submit_join_as_subcommunity_request(
+    parent_community_uuid, community_uuid, uow, comment=None
+):
+    """Create and submit a SubCommunityInvitation request."""
+    type_ = current_request_type_registry.lookup(ZenodoSubCommunityInvitationRequest.type_id)
+    creator = ResolverRegistry.resolve_entity_proxy(
+        {"community": parent_community_uuid}
+    ).resolve()
+    receiver = ResolverRegistry.resolve_entity_proxy(
+        {"community": community_uuid}
+    ).resolve()
+    expires_at = datetime.utcnow() + timedelta(days=30)
+
+    request_item = current_requests_service.create(
+        system_identity,
+        {},
+        type_,
+        receiver,
+        creator=creator,
+        expires_at=expires_at,
+        uow=uow,
+    )
+
     return current_requests_service.execute_action(
         system_identity, request_item._request.id, "submit", data=comment, uow=uow
     )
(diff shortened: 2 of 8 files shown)


site/zenodo_rdm/legacy/requests/utils.py
```diff
@@ -107,3 +108,32 @@ def submit_community_manage_record_request(user_id, uow, comment=None):
...
+
+    request_item = current_requests_service.create(
+        system_identity,
+        {},
+        type_,
+        receiver,
+        creator=creator,
+        expires_at=expires_at,
+        uow=uow,
+    )
```
slint, line 135:
```suggestion
    parent_community = community_service.record_cls.pid.resolve(parent_community_id)
    child_community = community_service.record_cls.pid.resolve(child_community_id)

    expires_at = datetime.utcnow() + timedelta(days=30)

    request_item = current_requests_service.create(
        system_identity,
        {},
        type_,
        creator=parent_community,
        receiver=child_community,
        expires_at=expires_at,
        uow=uow,
    )
```

nit: initializing with the resolver and resolving is a bit of a roundtrip vs. just resolving the communities. otherwise, some renames for being explicit/descriptive about who is the creator/receiver.

site/zenodo_rdm/subcommunities/request.py
```diff
@@ -113,8 +124,180 @@ class ZenodoSubCommunityRequest(SubCommunityRequest):
...
+
+class SubcommunityInvitationCreateAction(actions.CreateAction):
+    """Represents an accept action used to accept a subcommunity."""
+
+    def execute(self, identity, uow):
+        """Execute approve action."""
+        parent = self.request.created_by.resolve()
+        if not parent.children.allow:
+            raise ValidationError("Assigned parent is not allowed to be a parent.")
+        super().execute(identity, uow)
```
slint, line 143: minor: will be merged with the "Submit" action to become "CreateAndSubmit". it also looks like it's not necessary, since it's the same as the CreateAction from invenio-communities for the subcommunity invitation

site/zenodo_rdm/subcommunities/request.py
```diff
@@ -113,8 +124,180 @@ class ZenodoSubCommunityRequest(SubCommunityRequest):
...
+    """Represents an accept action used to accept a subcommunity.
+
+    Zenodo re-implementation of the accept action, to also move the records.
+    """
+
+    def _get_community_records(self, community_id):
+        """Get the records of a community."""
+        return current_community_records_service.search(
+            system_identity, community_id=community_id
+        )
```
slint, line 156: major/shelve: this method only returns the first 25 results from the record search, so the code below doesn't "bulk add" all the records of the community.

This actually is a bug even in our subcommunity inclusion request logic above, which means that when we accept subcommunity requests, this doesn't add ALL the records of the community to the parent community. 

It's relatively low impact though, since most communities don't have that many records (or probably have no records at all, since they've just being created). For the new request type though, this will not be the case, so this has to be fixed relatively quick.

site/zenodo_rdm/subcommunities/request.py
```diff
@@ -113,8 +124,180 @@ class ZenodoSubCommunityRequest(SubCommunityRequest):
...
+    """Submit action."""
+
+    def execute(self, identity, uow):
+        """Execute the submit action."""
+        self.request["title"] = "Invitation to join the EU Open Research Repository"
+
+        # example: "May 11, 2024"
+        expires_at = self.request.expires_at.strftime("%B %d, %Y")
+        NAME = "TODO"
+        ACRONYMS = "TODO"
```
slint, line 181: minor/shelve: these will probably be part of the data/payload of the request, so we can easily pass them as parameters when bulk creating the requests.

site/zenodo_rdm/subcommunities/request.py
```diff
@@ -113,8 +124,180 @@ class ZenodoSubCommunityRequest(SubCommunityRequest):
...
+
+        # Move records
+        records = self._get_community_records(child)
+        current_rdm_records.record_communities_service.bulk_add(
+            system_identity, parent, (x["id"] for x in records), uow=uow
+        )
+
+        current_communities.service.bulk_update_parent(
+            system_identity, [child], parent_id=parent, uow=uow
+        )
```
slint, line 247: minor: this can be DRYed out a bit, since it looks like there's some repetition (that can also be faulty, as we saw above).

site/zenodo_rdm/subcommunities/request.py
```diff
@@ -113,8 +124,180 @@ class ZenodoSubCommunityRequest(SubCommunityRequest):
...
+        type_,
+        receiver,
+        creator=creator,
+        expires_at=expires_at,
+        uow=uow,
+    )
+
+    return current_requests_service.execute_action(
+        system_identity, request_item._request.id, "submit", data=comment, uow=uow
+    )
```
slint, line 303: minor: I think this is a duplicate of the function in `utils.py`?

site/zenodo_rdm/subcommunities/request.py
```diff
@@ -6,21 +6,35 @@
...
+from invenio_rdm_records.proxies import (current_community_records_service,
+                                         current_rdm_records)
 from invenio_records_resources.services.uow import RecordCommitOp
 from invenio_requests.customizations import actions
 from invenio_requests.customizations.event_types import CommentEventType
 from invenio_requests.proxies import current_events_service
 from marshmallow import fields
+from werkzeug.local import LocalProxy
+
+community_service = LocalProxy(lambda: current_communities.service)
```
kpsherva, line 27: just out of curiosity, why additional localProxy wrap is needed?

site/zenodo_rdm/subcommunities/request.py
```diff
@@ -113,8 +117,111 @@ class ZenodoSubCommunityRequest(SubCommunityRequest):
 
     available_actions = {
         "delete": actions.DeleteAction,
-        "create": SubcommunityCreateAction,
         "cancel": actions.CancelAction,
+        # Custom implemented actions
```
kpsherva, line 121: what do you mean by "custom implemented actions"? custom = only present in zenodo?

site/zenodo_rdm/subcommunities/request.py
```diff
@@ -113,8 +117,111 @@ class ZenodoSubCommunityRequest(SubCommunityRequest):
...
+
+    def execute(self, identity, uow):
+        """Execute approve action."""
+        self.request["title"] = "Invitation to join the EU Open Research Repository"
+
+        # example: "May 11, 2024"
+        expires_at = self.request.expires_at.strftime("%B %d, %Y")
+        NAME = "TODO"
+        ACRONYMS = "TODO"
+        self.request["description"] = (
```
kpsherva, line 139: shouldn't this be in a template?

site/zenodo_rdm/subcommunities/request.py
```diff
@@ -113,8 +117,111 @@ class ZenodoSubCommunityRequest(SubCommunityRequest):
...
+
+class SubcommunityInvitationExpireAction(actions.ExpireAction):
+    """Expire action."""
+
+    def execute(self, identity, uow):
+        """Execute expire action."""
+        child = self.request.receiver.resolve().id
+        parent = self.request.created_by.resolve().id
+
+        _add_community_records(child, parent, uow)
```
kpsherva, line 199: does it mean you add records to parent community when invitation expires? Since I don't know the requirements it seems a bit counter intuitive.

carlinmack: yep, by default they will be included if the request expires. The thinking is that this is mutually beneficial and the request would only be declined if there has been a mistake

site/zenodo_rdm/subcommunities/request.py
carlinmack: Custom meaning that they aren't using the default `actions.Action`. Comment was added to match invenio-communities 
https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/subcommunities/services/request.py#L70

site/zenodo_rdm/subcommunities/request.py
carlinmack: community_service isn't used so I think this is a mistake, thanks!

scripts/sleep.py
carlinmack, line 1: 😴

slint: Hh no, my infamous sleep.py script for deployment shenanigans 😱

carlinmack: 🛌

PR #1068, opened 2024-11-18
Bump deps and various fixes

slint:
- **scripts: fix hashbang for "run_script.py"**
- **docker: remove obsolete "version" field**
- **config: fix parent DOI link**
- **installation: bump dependencies**

### DIFF
--- a/docker-compose.full.yml
+++ b/docker-compose.full.yml
@@ -1,4 +1,3 @@
-version: '2.2'
 services:
   cache:
     extends:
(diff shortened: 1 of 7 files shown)


docker-compose.full.yml
```diff
@@ -1,4 +1,3 @@
-version: '2.2'
```
slint, line 1: `version` is deprecated (and prints a warning for a while now)

PR #1072, opened 2024-11-26
curation: init module; add EURecordCurator

yashlamba:
Closes https://github.com/zenodo/ops/issues/505


site/zenodo_rdm/curation/tasks.py
```diff
@@ -0,0 +1,73 @@
...
+
+    query = dsl.Q(
+        "bool",
+        must=[
+            dsl.Q("term", **{"metadata.funding.funder.id": "00k4n6c32"}),
+            dsl.Q("term", **{"is_deleted": False}),
+            dsl.Q(
+                "range",
+                created={
+                    "lte": (datetime.now() - timedelta(days=30)).isoformat(),
```
slint, line 35:
```suggestion
                    "lte": (datetime.utcnow() - timedelta(days=30)).isoformat(),
```

or the "new way" using `datetime.timezone.utc`, since it's deprecated

site/zenodo_rdm/curation/tasks.py
```diff
@@ -0,0 +1,73 @@
...
+                "range",
+                updated={
+                    "gte": datetime.fromisoformat(since).isoformat(),
+                },
+            ),
+        ],
+        must_not=[
+            dsl.Q(
+                "term",
+                **{"parent.communities.ids": current_app.config.get("EU_COMMUNITY_ID")},
```
slint, line 48:
```suggestion
                **{"parent.communities.ids": current_app.config.get("EU_COMMUNITY_ID")},
```

This has to be the `Community.id` which is a UUID. If we're using the slug elsewhere it should be aligned (so we don't have to have so much configuration when deploying)

site/zenodo_rdm/curation/curators.py
```diff
@@ -0,0 +1,95 @@
...
+            rule_score = current_curation.scores.get(rule)
+            if isinstance(rule_score, int):
+                score += rule_score if result else 0
+            elif isinstance(rule_score, bool):
+                if result:
+                    return rule_score
+                else:
+                    continue
+            else:
+                raise ValueError("Unsupported score type configured.")
```
slint, line 72: This will also handle `None` if the rule fails above and `raise_exc=False`

site/zenodo_rdm/curation/tasks.py
```diff
@@ -0,0 +1,73 @@
...
+            )
+        ],
+    )
+    search = records_service.create_search(
```
slint, line 52: minor/comment: make sure we also check scenarios with different versions of the record being published

site/zenodo_rdm/curation/curators.py
```diff
@@ -0,0 +1,95 @@
...
+    def rules(self):
+        """Get rules to run from config."""
+        return current_app.config.get("CURATION_EU_RULES", {})
+
+    def _post_run(self, record, result):
+        """Actions to take after run."""
+        if self.dry:
+            current_app.logger.info(
+                f"Processed record ID: {record.pid.pid_value}", result
+            )  # TODO use error? Or should we log from the task
```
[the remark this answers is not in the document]
yashlamba: For this, my thought was should we log the record IDs just once here: https://github.com/zenodo/zenodo-rdm/blob/e3e4366fa067b2ab962324ac899184fa46e331ee/site/zenodo_rdm/curation/tasks.py#L70-L73

Instead of doing individual error logs which will also show up in the pod.

site/zenodo_rdm/curation/tasks.py
yashlamba: In the places I'm using it works with the UUID. I'll explicitly name the var UUID.

site/zenodo_rdm/curation/curators.py
yashlamba: Now that I think of it, raise_exc should only be a param of run

site/zenodo_rdm/curation/curators.py
```diff
@@ -0,0 +1,95 @@
...
+            if isinstance(rule_score, int):
+                score += rule_score if result else 0
+            elif isinstance(rule_score, bool):
+                if result:
+                    return rule_score
+                else:
+                    continue
+            else:
+                raise ValueError("Unsupported score type configured.")
+        return score >= current_app.config.get("CURATION_EU_CURATION_THRESHOLD")
```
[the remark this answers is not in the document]
yashlamba: Did it like scores, I think it'll be used often when/if we have other curators

site/zenodo_rdm/curation/curators.py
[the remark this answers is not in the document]
yashlamba: Ok, I'll push this as error log for now since we were planning to implement a custom logging logic anyways

PR #1082, opened 2024-12-03
serializers: added dcat serializer

0einstein0:
Partially fixes: https://github.com/inveniosoftware/invenio-rdm-records/issues/1889

### DIFF
--- a/site/zenodo_rdm/serializers/dcat.py
+++ b/site/zenodo_rdm/serializers/dcat.py
@@ -1,0 +1,69 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2024 CERN.
+#
+# ZenodoRDM is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+"""Zenodo dcat serializer."""
+
+
+import idutils
+from datacite import schema43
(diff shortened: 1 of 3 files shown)


site/zenodo_rdm/serializers/dcat.py
```diff
@@ -0,0 +1,119 @@
...
+from invenio_rdm_records.resources.serializers.dcat import DCATSerializer
+from lxml import etree
+
+
+class ZenodoDCATSerializer(DCATSerializer):
+    """Zenodo DCAT Serializer."""
+
+    def __init__(self, **options):
+        """Constructor."""
+        super().__init__(**options)
```
slint, line 21:
```suggestion
```

minor: if not overridden, I guess it's not needed.

site/zenodo_rdm/serializers/dcat.py
slint, line 1: minor: looking at both of the new methods now, I think we can actually add them directly in `rdm-records` and thus avoid entirely overriding the serializer here. I thought there would be too many specific "EDMO"-isms or similar for the other subject types, but it looks like all the namespaces and elements pretty much already exist in DCAT...

You did too good of a job at it, so now you get a bit more work 🙃

slint (approved):

PR #1087, opened 2024-12-05
curation: add new rules for EU curation

yashlamba:


site/zenodo_rdm/curation/config.py
```diff
@@ -10,27 +10,45 @@
...
 
-CURATION_THRESHOLDS = {"EU_RECORDS_CURATION": 10}
+CURATION_THRESHOLDS = {"EU_RECORDS_CURATION": 15}
 """Threshold values for curators/rules."""
 
 
 CURATION_ENABLE_EU_CURATOR = False
 """Controls whether to dry run EU Curation."""
+
+CURATION_LOW_CONF_KEYWORDS_EU = []
```
slint, line 50: Do you know how big is the list of keywords? If it's more than 50 elements it can become a bit clunky to have in OpenShift config

site/zenodo_rdm/curation/rules.py
```diff
@@ -4,19 +4,22 @@
...
 def award_acronym_in_description(record):
     """Check if EU award name in record description."""
     award_service = current_service_registry.get("awards")
-    description = record.metadata["description"]
-    funding = record.metadata["funding"]
+    description = record.metadata.get("description")
+    if not description:
+        return False
 
+    funding = record.metadata["funding"]
```
slint, line 22:
```suggestion
    funding = record.metadata.get("funding", [])
```

minor: safety first :)

site/zenodo_rdm/curation/rules.py
```diff
@@ -4,19 +4,22 @@
...
     for f in funding:
         if f["funder"]["id"] == "00k4n6c32":
```
slint, line 24:
```suggestion
        if f["funder"].get("id") == "00k4n6c32":
```

minor: again, I know that we prefilter the data in the query, but in case we change the logic at some point it's good to be on the safe side

site/zenodo_rdm/curation/rules.py
```diff
@@ -56,3 +59,56 @@ def test_phrases_in_record(record):
...
+def published_before_award_start(record):
+    """Check if published before award start date."""
+    award_service = current_service_registry.get("awards")
+
+    for f in record.metadata["funding"]:
+        if f["funder"]["id"] == "00k4n6c32":
+            if award_id := f.get("award", {}).get("id"):
+                award = award_service.record_cls.pid.resolve(award_id)
+                if award.get("start_date") and (
+                    record.created < datetime.fromisoformat(award.get("start_date"))
```
slint, line 73: nit: we can also use `arrow.get(...)` since it's a bit shorter and consistent across the codebase

site/zenodo_rdm/curation/rules.py
```diff
@@ -56,3 +59,56 @@ def test_phrases_in_record(record):
...
     return False
+
+
+def published_before_award_start(record):
+    """Check if published before award start date."""
+    award_service = current_service_registry.get("awards")
+
+    for f in record.metadata["funding"]:
+        if f["funder"]["id"] == "00k4n6c32":
+            if award_id := f.get("award", {}).get("id"):
```
slint, line 70: 🦭 :tada:

slint (approved):

site/zenodo_rdm/curation/config.py
yashlamba: expected to be 5-10

Issue #1095, opened 2024-12-11
Signup page: Add info box to ask people to use institutional email address


0einstein0:
![Screenshot 2024-12-13 at 14 24 16](https://github.com/user-attachments/assets/512c91ef-6b10-4dc3-84d8-58a293198f86)

PR #1101, opened 2024-12-13
register: added info box for institutional email

0einstein0:
Closes Issue https://github.com/zenodo/zenodo-rdm/issues/1095


templates/semantic-ui/invenio_accounts/register_user.html
```diff
@@ -52,7 +62,10 @@ <h1 class="ui inverted header text-center">Research. Shared! Sign up today</h1>
       {{ form.hidden_tag() }}
 
       {%- block registration_form_fields scoped %}
-      {{ render_field(form.email, icon="user icon", autofocus=True, errormsg=False) }}
+        {{ render_field(form.email, icon="user icon", autofocus=True, errormsg=False) }}
+        {{ render_field_info(_('Please use an institutional email address'), _('An institutional email address helps us verify your account and prevents our spam filtering system from automatically blocking your account.')) }}
```
slint, line 66: @lnielsen is that the final text we want to have on the tooltip? I don't remember if we decided on something else.

assets/less/zenodo-rdm/modules/popup.overrides
```diff
@@ -48,3 +48,12 @@
...
 }
+
+.field-info{
+  padding: 0.4em !important;
+  margin-top: -0.7em !important;
+  [data-tooltip]::after {
+    width: 25vw;
+    white-space: break-spaces;
+  }
+}
```
ptamarit, line 59: Wondering if this is the right file to put this CSS rule.

templates/semantic-ui/invenio_accounts/register_user.html
ptamarit: @0einstein0 do you know if we can go ahead with this specific text?

ptamarit (approved):

sakshamarora1 (approved):

lnielsen (approved):

templates/semantic-ui/invenio_accounts/register_user.html
ptamarit: Lars confirmed it looks fine.

assets/less/zenodo-rdm/modules/popup.overrides
ptamarit: Actually, there's a `[data-tooltip]::after` just above this addition, so it does make sense.

ptamarit:
Force pushed small whitespace formatting.

PR #1104, opened 2024-12-17
curation: add request checks; award acronym in community check

yashlamba:


site/zenodo_rdm/curation/rules.py
```diff
@@ -157,3 +162,87 @@ def award_acronym_in_additional_description(record):
...
+
+    if comm_text:
+        award_service = current_service_registry.get("awards")
+        funding = record.metadata.get("funding", [])
+        for f in funding:
+            if f["funder"].get("id") == "00k4n6c32":
+                if award_id := f.get("award", {}).get("id"):
+                    award = award_service.record_cls.pid.resolve(award_id)
+                    if award.get("acronym") and (
+                        award.get("acronym").lower() in comm_text.lower()
```
slint, line 245: minor: I would do some form of word-splitting here to be safer with the acronym match. Otherwise a project with acronym `SHINE` would also be matched in the text "...this approach out**shine**s all other methods...". 

Maybe the same thing should be applied in other rules that check for the acronym?

slint (approved):
LGTM, some minor comments on safer checks

slint (approved):

PR #1110, opened 2024-12-19
curation: tasks: send results in email

yashlamba:


site/zenodo_rdm/curation/tasks.py
```diff
@@ -10,12 +10,39 @@
...
+Records Approved: {approved}
+Record IDs Approved:
+{records_moved}
+"""
+
+
+def _send_result_email(content):
+    """Send curation result as email."""
+    subject = f"EU Record Curation Processesed {datetime.now().date()}"
+    body = RESULT_EMAIL_BODY.format(finished_at=datetime.now(timezone.utc), **content)
```
slint, line 36:
```suggestion
    body = RESULT_EMAIL_BODY.format(finished_at=datetime.now(timezone.utc).isoformat(), **content)
```

nit: not sure about the rest of the context if it's `str`-able

yashlamba: I tested all of it on my local, seems to be working fine for everything.

PR #1117, opened 2025-01-25
Create devcontainer.json

khma93:


slint:
Thank you for the contribution, but we currently not in the process of adding devcontainers support to the repository. If and when that is the case we would want to very clearly have all the features and dev workflows that are described in https://inveniordm.docs.cern.ch/develop/, which will probably require a decent amount of additional configurations.

PR #1119, opened 2025-02-03
Consistent capitalisation and remove invalid attribute

carlinmack:
1. Capitalised the first word always after —
2. Removed full stop at end of sentence as it is unnecessary
3. I know I shouldn't put this in the same PR (I used the web UI sorry), but removed the `checked` attributed as it is invalid on `type=hidden` elements. I did test that you can still subscribe to the mailing list when `checked` is removed

image

### DIFF
--- a/templates/semantic-ui/zenodo_rdm/frontpage.html
+++ b/templates/semantic-ui/zenodo_rdm/frontpage.html
@@ -48,25 +48,25 @@
     <div class="ui segment bottom attached rdm-sidebar">
       <ul class="m-0">
         <li>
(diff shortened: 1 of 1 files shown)


slint:
Test comment

ptamarit (approved):
I'm not convinced that uppercase and no dot is the best option, but at least this makes it consistent.

carlinmack:
Wikipedia does endorse the full stop, for me it's not necessary in this context

https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style/Lists#Format_list_items_consistently

image

Think I prefer lowercase but I wanted to follow the status quo

jrcastro2 (approved):
Peer reviewed with @0einstein0

Issue #1131, opened 2025-03-19
checks: setup/fixtures


slint:
We need a small script (or even better fixtures) that sets up `CheckConfig` for the EU community and a sub-community.

The same script will also be used for creating the checks on Sandbox and Production as part of the deployment to:

- Create a metadata check config with all the rules for the EU community
- Create for each sub-community a metadata check config with a rule that requires funding from the community's award/project

---

### Current rules configuration

```json
{
    "rules": [
        {
            "id": "access:open/publication",
            "title": "Open Access Publication",
            "message": "Publication articles must be Open Access",
            "description": "The EU curation policy requires publication articles must be Open Access",
            "level": "error",
            "condition": {
                "type": "comparison",
                "left": {
                    "type": "field",
                    "path": "metadata.resource_type.id"
                },
                "operator": "==",
                "right": "publication-article"
            },
            "checks": [
                {
                    "type": "comparison",
                    "left": {
                        "type": "field",
                        "path": "access.files"
                    },
                    "operator": "==",
                    "right": "public"
                }
            ]
        },
        {
            "id": "journal:title/publication",
            "title": "Journal Information",
            "message": "Publication articles must state the journal it was published in",
            "description": "The EU curation policy requires publication articles must state the journal it was published in",
            "level": "error",
            "condition": {
                "type": "comparison",
                "left": {
                    "type": "field",
                    "path": "metadata.resource_type.id"
                },
                "operator": "==",
                "right": "publication-article"
            },
            "checks": [
                {
                    "type": "comparison",
                    "left": {
                        "type": "field",
                        "path": "custom_fields.journal:journal.title"
                    },
                    "operator": "!=",
                    "right": ""
                }
            ]
        },
        {
            "id": "license:exists",
            "title": "Record license",
            "message": "All records must have a license",
            "description": "The EU curation policy requires all records to have a license",
            "level": "error",
            "checks": [
                {
                    "type": "list",
                    "operator": "exists",
                    "path": "metadata.rights",
                    "predicate": {}
                }
            ]
        },
        {
            "id": "license:cc-by/publication",
            "title": "Journal Article License",
            "message": "Journal article should have a CC-BY license",
            "description": "The EU curation policy recommends journal articles to have a CC-BY license",
            "level": "info",
            "condition": {
                "type": "logical",
                "operator": "and",
                "expressions": [
                    {
                        "type": "comparison",
                        "left": {
                            "type": "field",
                            "path": "metadata.resource_type.id"
[...]

PR #1137, opened 2025-03-21
scripts: add checks seeding for EU communities

slint:
- Closes https://github.com/zenodo/zenodo-rdm/issues/1131


ptamarit:
One extra commit in master is removing `access:open/publication`.
See https://github.com/inveniosoftware/invenio-rdm-records/pull/1975 for the rule definition to add back.

PR #1145, opened 2025-04-01
templates: removed the files macros

0einstein0:


slint (approved):
> Begone foul template!

Issue #1169, opened 2025-04-14
metadata: thesis custom field migration


slint:
- [Old fields structure](https://github.com/inveniosoftware/invenio-rdm-records/blob/v17.6.0/invenio_rdm_records/contrib/thesis/custom_fields.py) 
- [New fields structure](https://github.com/inveniosoftware/invenio-rdm-records/blob/ebca561cb385eb85d1955f83fc2969a68f9e08d8/invenio_rdm_records/contrib/thesis/custom_fields.py)
- [Search query](https://zenodo.org/search?q=_exists_%3Acustom_fields.thesis%5C%3Auniversity&f=allversions%3Atrue&l=list&p=1&s=10&sort=bestmatch) to match existing records with `thesis:university` field.
- CodiMD: http://codimd.web.cern.ch/PPXsabZ6R2apIbxstw2SrA?view

---

### Code/UI updates

- The Legacy REST API [accepts the field](https://github.com/zenodo/zenodo-rdm/blob/93873f7619a7114d5225ce97f0545e2ef94ff25b/site/zenodo_rdm/legacy/deserializers/schemas.py#L129). It should first put the values in both fields, and after removal and data migration only to the new one
- The UI displays this field as ["Awarding University"](https://zenodo.org/records/12797117). Same as above, it should be able to handle both, but prefer the new value, i.e. `thesis:thesis.university` if both are available. This code is [in app-rdm](https://github.com/inveniosoftware/invenio-app-rdm/blob/8bb1652cf76d1eb63dce8c60d5f83ca7434dcc7b/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/details/side_bar/details.html#L29).
  - After closer inspection, it looks like this relies on the `publishing_information` UI serialized field. We should make sure that `thesis:thesis.univcersity` falls back to `thesis:university`.

### Migration steps

We need a recipe with steps on how to do the following:

1. Setup the new custom field on the backend and mappings only
    - `invenio rdm-records custom-fields init -f thesis:thesis`
    - Make sure the we still have both the `tehsis:university` and `thesis:thesis` fields in the custom fields config
    - Make sure we only have the `thesis:university` in the UI config
2. Copy all the data from the records that have a `thesis:university` field to the new `thesis:thesis.unviersity` field and reindex the records
3. Switch the UI config from the old field to the new one
4. Remove all the data from the old field
5. Remove all config of the old field

PR #1171, opened 2025-04-16
custom_fields: added old thesis field for migration

0einstein0:
Closes Issue https://github.com/zenodo/zenodo-rdm/issues/1169


slint (commented):
Some minor comments. I'm not sure about the tests... I think we need to have closer look to debug it...

PR #1175, opened 2025-04-24
celery: create indexing queue and add more tasks to low

carlinmack:
1. Creating indexing queue to reduce load on the default queue
2. Add all clean up tasks to low queue

### DIFF
--- a/invenio.cfg
+++ b/invenio.cfg
@@ -117,7 +117,12 @@ RDM_COMMUNITIES_ROUTES["community-home"] = (
     communities_home,
 )
 
-ZENODO_ENV = os.environ.get("ZENODO_ENV", "production").lower()
+ZENODO_ENV = os.environ.get("ZENODO_ENV", "local").lower()
+VALID_ENVS = ["local", "dev", "qa", "production"]
+if ZENODO_ENV not in VALID_ENVS:
+    raise ValueError(f"ZENODO_ENV is set to an invalid value: {ZENODO_ENV}\nExpected one of {VALID_ENVS}")
+
+IS_LOCAL_DEV = ZENODO_ENV == "local"
 
 
 # Flask-SQLAlchemy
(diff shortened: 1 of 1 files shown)


invenio.cfg
```diff
@@ -238,9 +238,41 @@ FILES_REST_CHECKSUM_VERIFICATION_URI_PREFIXES = {
...
+    "invenio_users_resources.services.users.tasks.unindex_users": {"queue": "indexing"},
     # Low
+    "invenio_accounts.tasks.clean_session_table": {"queue": "low"},
+    "invenio_files_rest.tasks.clear_orphaned_files": {"queue": "low"},
+    "invenio_files_rest.tasks.remove_expired_multipartobjects": {"queue": "low"},
+    "invenio_files_rest.tasks.remove_file_data": {"queue": "low"},
     "invenio_files_rest.tasks.verify_checksum": {"queue": "low"},
+    "invenio_rdm_records.requests.user_moderation.tasks.user_block_cleanup": {
+        "queue": "low"
+    },
```
carlinmack, line 269: maybe blocking cleanup shouldn't be in low actually

carlinmack: put in spam queue

invenio.cfg
```diff
@@ -238,9 +238,41 @@ FILES_REST_CHECKSUM_VERIFICATION_URI_PREFIXES = {
 }
 
 CELERY_TASK_ROUTES = {
+    # Indexing
```
carlinmack, line 241: add If not deployed environment

invenio.cfg
```diff
@@ -117,7 +117,8 @@ RDM_COMMUNITIES_ROUTES["community-home"] = (
     communities_home,
 )
 
-ZENODO_ENV = os.environ.get("ZENODO_ENV", "production").lower()
+ZENODO_ENV = os.environ.get("ZENODO_ENV", "local").lower()
+IS_DEPLOYED = ZENODO_ENV in ["production", "qa", "dev"]
```
slint, line 121:
```suggestion
IS_LOCAL_DEV = ZENODO_ENV == "local"
```

The reasoning is that if e.g. one makes a mistake and sets `ZENODO_ENV=prod` on production, we would still get "some" sort of production behavior (i.e. tasks getting distributed to queues), since `IS_LOCAL_DEV` would be `False`.

slint (approved):

invenio.cfg
carlinmack: from what you say I feel like it's more important that line 120 has default = `production`? 

Additionally I don't think `IS_LOCAL_DEV` would ever be true in this case

slint: For local development, nobody sets `ZENODO_ENV` in their environment (and I wouldn't start enforcing anything like that). If we have `ZENODO_ENV = os.environ.get("ZENODO_ENV", "local").lower()`, then by default this would be set to `"local"` and thus `IS_LOCAL_DEV` would be set to `True`

carlinmack: okay yeah I get it, I just mean that maybe we care about if the ENV isn't set on prod and therefore we shouldn't change the default

carlinmack: anyway I've pushed the above

slint (approved):

invenio.cfg
slint: Agree regarding being defensive, in which case we should also add something like:

```python
VALID_ENVS = ["local", "dev", "qa", "production"]
if ZENODO_ENV not in VALID_ENVS:
    raise ValueError(f"ZENODO_ENV is set to an invalid value: {ZENODO_ENV}\nExpected one of {VALID_ENVS}"
``` 

I was unhappy with `IS_DEPLOYED = ZENODO_ENV in ["production", "qa", "dev"]`, because in case of a typo, you end up with `IS_DEPLOYED = False` in `prod`

slint (approved):

ptamarit:
I just added `worker-indexing` on QA and PROD (see OpenShift export).

PR #1180, opened 2025-05-08
File format checks

slint:
- Closes https://github.com/inveniosoftware/invenio-checks/issues/27
- **WIP checks: configure file format checks**


scripts/ec_create_checks.py
```diff
@@ -593,5 +579,46 @@ def create_eu_checks():
...
+        db.session.add(check_config)
+    db.session.commit()
+    print(f"Metadata checks created/updated successfully for community {eu_comm.slug}.")
+
+
+def create_file_format_checks(eu_comm):
+    existing_check = CheckConfig.query.filter_by(
+        community_id=eu_comm.id, check_id="file_formats"
+    ).one_or_none()
+    if not existing_check:
```
carlinmack, line 605: why not do the if ... else like above?

PR #1183, opened 2025-05-09
checks: create help msg override

carlinmack:
Closes https://github.com/inveniosoftware/invenio-checks/issues/27


slint:
Merged, since it will only take effect after updating to latest `invenio-checks`

PR #1188, opened 2025-05-27
checks: added message box for draft in progress

0einstein0:
Screenshot 2025-05-27 at 16 09 19

### DIFF
--- a/site/zenodo_rdm/templates/semantic-ui/invenio_checks/requests/help_message.html
+++ b/site/zenodo_rdm/templates/semantic-ui/invenio_checks/requests/help_message.html
@@ -7,6 +7,12 @@ Invenio is free software; you can redistribute it and/or modify it
 under the terms of the MIT License; see LICENSE file for more details.
 #}
 
+{% if has_draft %}
+<div class="ui warning message">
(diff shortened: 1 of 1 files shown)


site/zenodo_rdm/templates/semantic-ui/invenio_checks/requests/help_message.html
```diff
@@ -7,6 +7,12 @@
 under the terms of the MIT License; see LICENSE file for more details.
 #}
 
+{% if has_draft %}
+<div class="ui warning message">
+    <div class="header">Note: Draft in progress</div>
+    <p>The checks and feedback are based on the current draft of the record. Any recent edits may not yet appear in the published version.</p>
+</div>
```
slint, line 14:
```suggestion
    <div class="header"><i class="attention icon"></i> Draft editing in progress</div>
    <p>You're viewing checks and feedback results for unpublished changes.</p>
</div>
```

Then maybe we could also add another line if we know there is also a published record? Something like:


```jinja
{% if is_published %} The published version available to users may differ from what's shown here.{% endif %}
```

PR #1199, opened 2025-06-23
Switch to uv

slint:
- **assets: remove trailing slash from JS import**
  * This causes issues when building assets with rspack.
  

- **global: switch from pipenv to uv**

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,15 +52,9 @@ for further installation options.
 
 ### Update dependencies
 
-To update dependencies you need to run `pipenv lock` in the target deployment
-environment:
 
-```shell
-# Run the container with x86_64 architecture
-docker run -it --platform="linux/amd64" --rm -v $(pwd):/app \
-    registry.cern.ch/inveniosoftware/almalinux:1
+To update dependencies you need to:
 
-# Inside the container update the Pipfile.lock
-[root@3954486e4a37]# cd /app
-[root@3954486e4a37]# pipenv lock
-```
+1. Run `invenio-cli packages lock`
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -43,29 +41,29 @@ jobs:
...
 
-      - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v5
+      - name: Cache Docker images
+        uses: ScribeMD/docker-cache@fb28c93772363301b8d0a6072ce850224b73f74e  # 0.5.0
         with:
-          python-version: ${{ matrix.python-version }}
-          cache: pip
-          cache-dependency-path: Pipfile.lock
+          key: docker-${{ runner.os }}-${{ hashFiles('docker-services.yml') }}
```
slint, line 47: Not entirely correct to use the `docker-service.yml` file, since the pulled Docker images are the ones that come baked-in the fetched release of `docker-services-cli`... But it's good enough and we should probably keep the versions synced

.github/workflows/tests.yml
```diff
@@ -43,29 +41,29 @@ jobs:
...
 
-          # We're pinning "pipenv" to the same version as in the deployment Docker image
-          pip install "pipenv==2023.12.1"
-          [ "${{ matrix.package-dir }}" == "site" ] && pipenv install --deploy --system
-          pip install "./legacy"
-          pip install "./${{ matrix.package-dir }}[$EXTRAS]"
+      - name: Install system dependencies
+        uses: awalsh128/cache-apt-pkgs-action@4c82c3ccdc1344ee11e9775dbdbdf43aa8a5614e  # v1.5.1
+        with:
+          packages: libkrb5-dev libvips-dev
```
slint, line 58: These took 1-2min to build before :eyes:

PR #1212, opened 2025-07-08
fix(jobs): since string cast

palkerecsenyi:
**Please release this PR at the same time as the changes in https://github.com/inveniosoftware/invenio-jobs/issues/87#issuecomment-3045453052.**

Closes https://github.com/inveniosoftware/invenio-jobs/issues/87

---

* Removed the deprecated utcnow() function in the curation/tasks.py file
and replaced it with the recommended new syntax.

* Remove the string cast for the `since` argument in the curation job

### DIFF
--- a/site/zenodo_rdm/curation/jobs.py
+++ b/site/zenodo_rdm/curation/jobs.py
@@ -24,4 +24,5 @@ class EURecordCuration(JobType):
 
     @classmethod
     def build_task_arguments(cls, job_obj, since=None, **kwargs):
+        """We only need the since argument."""
         return {"since": since or datetime.now(timezone.utc)}


site/zenodo_rdm/curation/jobs.py
```diff
@@ -30,6 +28,6 @@ def build_task_arguments(cls, job_obj, since=None, **kwargs):
         if since is None and job_obj.last_runs["success"]:
             since = job_obj.last_runs["success"].started_at
         else:
-            since = since or datetime.utcnow()
+            since = since or datetime.now(timezone.utc)
 
-        return {"since": str(since)}
+        return {"since": since}
```
kpsherva, line 33: I might have misunderstood the goal of normalising "since" parameter, but I thought we aimed to remove this return from all the child JobType class methods - and the idea was to return it by default in the equivalent private method, ping @ntarocco for confirmation

ntarocco: > I might have misunderstood the goal of normalising "since" parameter, but I thought we aimed to remove this return from all the child JobType class methods - and the idea was to return it by default in the equivalent private method, ping @ntarocco for confirmation

The goal was not to remove entirely the `since`, but only the casting to str. If, for some reason, the job requires the `since` param (because it has to compute some dates or be aware of it), then the job implementation will return it and inject in the celery task.

@palkerecsenyi it looks that [it is needed](https://github.com/zenodo/zenodo-rdm/blob/3627edef32c64b314a92dc81ab9c0bd69b2054f4/site/zenodo_rdm/curation/tasks.py#L107)!

PR #1241, opened 2025-09-08
Support form feedback changes and disable for record deletion

ptamarit:
This can be reviewed commit by commit:
1. Fix global error message for validation errors
    If the description was shorter than 20 characters, we would tell users to contact us via email
2. Show success message in React
    We were using Flask's `flash` and `redirect` and then were changing the location of the page.
    When the form was submitted via JavaScript, the API response would redirect to the homepage with the flash message in the HTML, but we were not showing this HTML content.
    We were then changing the location of the page to the homepage, but the flash message would not be there anymore, since this was the 2nd loading the same page.
    Partially fixes zenodo/ops#522
3. Disable the form for record deletion
    Changes the message and disables the form to avoid getting tickets for record deletion.


site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
```diff
@@ -133,7 +137,17 @@ class SupportForm extends Component {
...
     const sysInfo = `Browser: ${userBrowser} Operating System: ${userPlatform}`;
-    const { fileErrorMessage, rejectedFiles, loading, errorStatus } = this.state;
+    const { fileErrorMessage, rejectedFiles, loading, errorStatus, successMessage } =
+      this.state;
+
+    if (successMessage) {
+      return (
+        <Message success>
+          <MessageHeader>Request created</MessageHeader>
+          <p>{successMessage}</p>
```
slint, line 160: nit/shelve: I remember someone mentioned they would like to be able to see on a successful request what kind of information they submitted. Would there be a way to disable/grey-out the form or include in this message a display of the sent message?

Not critical, so we can shelve or skip (I remember we discussed this also briefly IRL, but forgot the conclusion 😅)

slint (approved):

site/zenodo_rdm/assets/semantic-ui/js/zenodo_rdm/src/support/components/SupportForm.js
ptamarit: Done, I'm including the category, subject and message in the response and show it in the success message box:
image

PR #1280, opened 2025-11-24
Override file modification to add link to support AND update max page size limits to 100 for auth, 25 for guest

slint:
CleanShot 2025-11-24 at 16 26 01@2x

### DIFF
--- a/site/tests/test_search.py
+++ b/site/tests/test_search.py
@@ -1,0 +1,61 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2025 CERN.
+#
+# ZenodoRDM is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Test search API page size validation."""
+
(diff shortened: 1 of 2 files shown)


assets/js/components/community/overrides/FileModificationModalMessage.js
```diff
@@ -0,0 +1,37 @@
...
+// as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+import React from "react";
+import { ModalContent } from "semantic-ui-react";
+
+const FileModificationModalMessage = () => {
+  return (
+    <ModalContent>
+      <p>
+        If you need to modify files in a deposit that was published more than 30 days ago, please contact our <a target="_blank" href="/support">support line</a>.
```
carlinmack:
```suggestion
        If you need to modify files in a record that was published more than 30 days ago, please contact our <a target="_blank" href="/support">support line</a>.
```
I think deposit is a bit dated

PR #1348, opened 2026-04-16
feat(docker): upgrade images to OpenSearch 3.4.0 and PostgreSQL 15.15

ptamarit:


docker-services.yml
```diff
@@ -34,7 +34,7 @@ services:
     ports:
       - "${DOCKER_SERVICES_IP_BIND:-127.0.0.1}:6379:6379"
   db:
-    image: postgres:14.15
+    image: postgres:15.15
```
slint, line 37: maybe 15.11 (IIRC), to match what we have on DBoD?

ptamarit: The latest minor is `15.15` (upgrade required before end of April).

PR #1355, opened 2026-04-26
feat(assets): switch to pnpm and rspack

slint:


run-js-linter.sh
```diff
@@ -5,18 +5,41 @@ GREEN='\033[0;32m'
...
 		*)
 			printf "Argument ${RED}$arg${NC} not supported\n"
-			exit;;
+			exit 1;;
 	esac
 done
 
+# Install lint deps in a project-local hidden dir to avoid peer-dep conflicts
+# with the root package.json (which is managed by `invenio-cli assets lock`).
+# NODE_PATH points eslint at the dir so it can resolve the shareable config.
```
slint: I tried a bunch of ways to avoid this crap here, but it was impossible (either failing in CI or locally). Some things tried (and failed or didn't like):

```bash
# 1. pnpm's npx
pnpm dlx --package=eslint@^8 --package=@inveniosoftware/eslint-config-invenio eslint ...
# deps end up in ~/Library/Caches/pnpm/dlx/<hash>/node_modules, but eslint
# resolves shareable configs from CWD walking up, so never finds them

# 2. hoisting
NPM_CONFIG_SHAMEFULLY_HOIST=true pnpm dlx --package=eslint@^8 --package=@inveniosoftware/eslint-config-invenio eslint ...
# flattens the dlx `node_modules` but doesn't fix CWD-based resolution.

# 3. one-liner
npx --yes -p eslint@^8 -p @inveniosoftware/eslint-config-invenio eslint ...
# same problem: deps go to ~/.npm/_npx/<hash>, not on the resolution path.

# 4. at project root
npm install --no-save --no-package-lock ...
# ERESOLVE: the root package.json (managed by `invenio-cli assets lock`)
# pulls in rspack peer-dep conflicts that block the install

# 5. pnpm dlx + NODE_PATH=$(dirname $(command -v eslint))/.. (inline)
# works, but ugly inline shell hack and re-downloads on every run.
```

.gitignore
```diff
@@ -67,6 +67,9 @@ logs/
 # Invenio-cli per machine file
 .invenio.private
 
+# JS linting temp dir
+.lint/
```
slint: See updated linting script for why we need this...

run-js-linter.sh
OliverGeneser: Yeah that's in fact very ugly.... 🤢

Hmm I'm not really a fan of generating the lint workspace inline (I also don't now how the runners handle the changing lock-file? What about committing a dedicated tools/lint workspace? And then explicitly caching the node_modules?

OliverGeneser (approved):
Works great in my local env 🎉

See comments about run-js-linter

run-js-linter.sh
slint: I think moving this inside the `site/` dir might also work... will give it a quick go

OliverGeneser (approved):

PR #1374, opened 2026-05-27
fix(serializers): use DataCite v4.5 encoder in XML output

slint:
* Makes sure that the DataCite XML serialization includes the `identifier` element.


ptamarit (approved):
Since we updated the JSON serializer and forgot the XML one, I'm wondering if we can find a better way to make sure we update the JSON and XML serializers at the same time next time. Maybe a variable grouping both declarations side-by-side?

```python
datacite_versions = {"serializer": DataCite45Schema, "encoder": schema45}
```

slint:
I split the two elements into constants, so hopefully they're close enough for the next change :)

PR #1386, opened 2026-06-15
ci: upgrade to actions using Node.js 24

ptamarit:
All our actions were versions running on Node.js 20, which is deprecated.
GitHub Actions are displaying the following message:

> Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: [...]. Actions will be forced to run with Node.js 24 by default starting June 16th, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

All these upgraded actions are using Node.js 24.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,22 +36,23 @@ jobs:
       EXTRAS: tests
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
 
       - name: Cache Docker images
         # We're using the fork by AndreKurait, which has bumped the "cache" action
-        uses: AndreKurait/docker-cache@0fe76702a40db986d9663c24954fc14c6a6031b7  # 0.6.0
+        uses: AndreKurait/docker-cache@7a3887908bdb97935395833df69b060cfcca0f7f  # 0.7.0
         with:
           key: docker-${{ runner.os }}-${{ hashFiles('docker-services.yml') }}
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v5
+        uses: astral-sh/setup-uv@v8.2.0
         with:
           python-version: ${{ matrix.python-version }}
           enable-cache: true
+          activate-environment: true
 
       - name: Install system dependencies
-        uses: awalsh128/cache-apt-pkgs-action@4c82c3ccdc1344ee11e9775dbdbdf43aa8a5614e  # v1.5.1
+        uses: awalsh128/cache-apt-pkgs-action@681749ae568c81c2037cb9185e38b709b261bd2f  # v1.6.1
         with:
           packages: libkrb5-dev libvips-dev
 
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -68,11 +68,11 @@ jobs:
...
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
       - name: Use Node.js v16.x
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@v6
         with:
-          node-version: v16.x
+          node-version: 24
```
ptamarit, line 76: There was no complaint about this runtime version, but I took this opportunity to align it with the versions used by the actions.

.github/workflows/tests.yml
```diff
@@ -36,22 +36,22 @@ jobs:
...
       - name: Cache Docker images
         # We're using the fork by AndreKurait, which has bumped the "cache" action
-        uses: AndreKurait/docker-cache@0fe76702a40db986d9663c24954fc14c6a6031b7  # 0.6.0
+        uses: AndreKurait/docker-cache@7a3887908bdb97935395833df69b060cfcca0f7f  # 0.7.0
         with:
           key: docker-${{ runner.os }}-${{ hashFiles('docker-services.yml') }}
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v5
+        uses: astral-sh/setup-uv@v8.2.0
```
ptamarit, line 48: `@v8` does not work due to [v8 starting to use immutable releases](https://github.com/astral-sh/setup-uv/releases/tag/v8.0.0).

.github/workflows/tests.yml
```diff
@@ -36,22 +36,23 @@ jobs:
...
         with:
           python-version: ${{ matrix.python-version }}
           enable-cache: true
+          activate-environment: true
```
ptamarit, line 52: [Since v6, the environment is not implicitly activated](https://github.com/astral-sh/setup-uv/releases/tag/v6.0.0) when setting up uv.

.github/workflows/tests.yml
slint: The only concern I have is that in the Docker image we build, we have Node v16.20.2 installed, so we would be testing against a different version.

At the same time, this is just linting/formatting, so I don't see a big problem with just bumping (and eventually once we upgrade to Python 3.14, etc. we'll anyways be on a newer version of Node).

slint (approved):
:shipit:

OliverGeneser (approved):

PR #1389, opened 2026-06-22
previewer: fix provide previewable_extensions attribute

ptamarit:
Before, images had a preview button because the default `iiif_simple` was registering image extensions.
Since inveniosoftware/invenio-previewer#243, we "Add the extension only if the previewer is in the preference list (it is enabled)".

We are fixing Zenodo by by providing the `previewable_extensions` attribute in our previewer, as all previewer should.

### DIFF
--- a/site/zenodo_rdm/previewer/image_previewer.py
+++ b/site/zenodo_rdm/previewer/image_previewer.py
@@ -7,6 +7,9 @@ from datetime import datetime, timedelta, timezone
 from os.path import splitext
 
 from flask import current_app, render_template
+from werkzeug.local import LocalProxy
+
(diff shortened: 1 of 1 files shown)


site/zenodo_rdm/previewer/image_previewer.py
```diff
@@ -7,6 +7,9 @@
 from os.path import splitext
 
 from flask import current_app, render_template
+from werkzeug.local import LocalProxy
+
+previewable_extensions = LocalProxy(lambda: current_app.config["IIIF_FORMATS"].keys())
```
ptamarit, line 12: Defined the same way as https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/records_ui/previewer/iiif_simple.py

slint (approved):

PR #1406, opened 2026-07-13
fix: 500 error on file deletion

OliverGeneser:


slint:
This, along with some other fixes (and tests to replicate) on the legacy files REST API are in https://github.com/zenodo/zenodo-rdm/pull/1408.

Especially for this issue, the fix is in https://github.com/zenodo/zenodo-rdm/pull/1408/changes/9d669097ac5e18eb5843dc3c097729e78d0c8adf (which is following also what we do upstream for 204 responses on `DELETE` requests).

PR #1407, opened 2026-07-15
feat(moderation): notify user by email on block

mairasalazar:
Add on_block_notify callback that sends an email to users who have been blocked by the automated spam-detection. The body includes the user id and the message includes a reply-to Support. The callback is skipped for human-triggered blocks (actor_id set).

### DIFF
--- a/site/zenodo_rdm/moderation/actions.py
+++ b/site/zenodo_rdm/moderation/actions.py
@@ -1,0 +1,41 @@
+# SPDX-FileCopyrightText: 2026 CERN
+# SPDX-License-Identifier: GPL-3.0-or-later
+"""RDM user moderation action."""
+
+from flask import current_app
+from flask_mail import Message
+from invenio_app_rdm.utils.files import render_email_from_context
+from invenio_users_resources.records.api import UserAggregate
+
+_BLOCK_NOTIFICATION_HTML = "zenodo_rdm/moderation/block_notification.html"
+_BLOCK_NOTIFICATION_TXT = "zenodo_rdm/moderation/block_notification.txt"
+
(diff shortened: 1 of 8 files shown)


site/zenodo_rdm/moderation/actions.py
```diff
@@ -0,0 +1,41 @@
...
+        body=render_email_from_context(_BLOCK_NOTIFICATION_TXT, context),
+        html=render_email_from_context(_BLOCK_NOTIFICATION_HTML, context),
+    )
+    try:
+        mail_ext.send(msg)
+    except Exception as e:
+        current_app.logger.exception(
+            "Block notification email report not sent. Error: {}",
+            extra={"user_id": user_id}
+        )
```
OliverGeneser: e is never used 😊

site/zenodo_rdm/moderation/actions.py
```diff
@@ -0,0 +1,41 @@
...
+        recipients=[user.email],
+        reply_to=current_app.config["MAIL_DEFAULT_SUPPORT"],
+        body=render_email_from_context(_BLOCK_NOTIFICATION_TXT, context),
+        html=render_email_from_context(_BLOCK_NOTIFICATION_HTML, context),
+    )
+    try:
+        mail_ext.send(msg)
+    except Exception as e:
+        current_app.logger.exception(
+            "Block notification email report not sent. Error: {}",
```
OliverGeneser: nit; as it's a internal error message I would prefer to follow "Failed to send block notification email"

OliverGeneser (approved):
lgtm

PR #1410, opened 2026-07-22
fix(doi): correct concept doi related identifier

ntarocco:
- the change to the DataCite JSON serializer parameter introduced with inveniosoftware/invenio-rdm-records/eaa4ba426a1a5f9ae192461105fb109183444b2a was not propagated to Zenodo, causing the error of minting the concept doi with the same related identifier as each record version

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -24,7 +24,7 @@ on:
 
 jobs:
   Python:
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-latest
     strategy:
       matrix:
           python-version: ["3.9"]
@@ -36,7 +36,7 @@ jobs:
       EXTRAS: tests
     steps:
       - name: Checkout
-        uses: actions/checkout@v6
+        uses: actions/checkout@v7
 
       - name: Cache Docker images
         # We're using the fork by AndreKurait, which has bumped the "cache" action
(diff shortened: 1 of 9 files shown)


site/zenodo_rdm/providers.py
```diff
@@ -200,7 +200,7 @@ def validate(self, record, identifier=None, provider=None, **kwargs):
     rdm_providers.DataCitePIDProvider(
         "datacite",
         client=rdm_providers.DataCiteClient("datacite", config_prefix="DATACITE"),
-        serializer=ZenodoDataciteJSONSerializer(schema_context={"is_parent": True}),
+        serializer=ZenodoDataciteJSONSerializer(is_parent=True),
```
ntarocco, line 203: This is the fix

slint (approved):

.github/workflows/tests.yml
```diff
@@ -24,7 +24,7 @@ on:
 
 jobs:
   Python:
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-latest
```
OliverGeneser, line 27: What's the reason for moving to a floating tag? 😊 
This make the CI less deterministic and might change Docker behavior, Python toolchain etc. but I guess we would also get the Github image security patches "potentially" faster 🤔

OliverGeneser (approved):
LGTM

.github/workflows/tests.yml
ntarocco: I don't think that a specific version is much more deterministic than `latest`: yes, the OS version is stable, however, I hope they install the latest patches and updates.
If we are worried about OS related issues, I believe that we should use the same OS that we use in prod. `latest` worked well in the past.

Feel free to revert it, no problem for me.

slint: For reference, we're [also using `ubuntu-latest` in Invenio CI](https://github.com/inveniosoftware/workflows/blob/443321c12bbf5d3866405e26d586cb42e6c2f185/.github/workflows/tests-python.yml#L31) for the last two weeks.

PR #1415, opened 2026-07-30
fix(quota): implement changes introduced in rdm-records

ntarocco:
⚠️ can be merged **ONLY** when [this](https://github.com/inveniosoftware/invenio-rdm-records/pull/2418) is merged, released and cherry-pick to latest rdm-records

### DIFF
--- a/site/zenodo_rdm/request_policies.py
+++ b/site/zenodo_rdm/request_policies.py
@@ -15,17 +15,19 @@ class QuotaIncreasePolicy(BasePolicy):
 
     def is_allowed(self, identity, record=None):
         """Only owners can increase the quota."""
-        if identity.user.verified_at:
-            if record:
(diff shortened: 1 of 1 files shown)


slint:
We had a discussion at some point in the chat to remove the verified conditional, since it causes more issues than it solves (e.g. folks not seeing the option in the deposit form). Does it make sense that we disable this for now, monitor if there's any bad mehavior, and then re-enable if needed?

ntarocco:
> We had a discussion at some point in the chat to remove the verified conditional, since it causes more issues than it solves (e.g. folks not seeing the option in the deposit form). Does it make sense that we disable this for now, monitor if there's any bad mehavior, and then re-enable if needed?

OK, done

slint (approved):

PR #1416, opened 2026-07-31
fix(checks): rerun the funding check only when its inputs change

slint:
`should_rerun` compares the input hash whatever the run's status. A pending or
running run carries the last completed hash, so an unrelated edit waits for the
run in flight instead of starting a second one.

Also schedule `cleanup_stale_check_runs`, which fails check runs whose worker
never came back.


mairasalazar (approved):
🚀

PR #1417, opened 2026-08-05
fix(checks): skip EOR funding check if no request

### LINKED ISSUE #1420: Improvements HZ+ checks
- [x] Avoid duplicate LLM calls for funding relevance check: https://github.com/zenodo/zenodo-rdm/pull/1417
- [x] Allow skipping checks: https://github.com/inveniosoftware/invenio-checks/pull/86 and https://github.com/zenodo/zenodo-rdm/pull/1418
    - [x] Skip rule results when the condition is not successful
    - [x] Avoid displaying empty check tabs when the check has filter_by_community = True
- [x] Add link to either user profile page or admin page for user in the community members check: https://github.com/zenodo/zenodo-rdm/pull/1419

mairasalazar:
If the check config for funding relevance is created for the parent community, each record or draft inclusion request  to a subcommunity will generate two check runs (one for the subcommunity and one for the parent).

To avoid performing the LLM call twice for both check runs unecessarily, for the check run for the parent config, we check if there exists an open or accepted request with the parent community as the receiver. If not, we short-circuit the check run. Otherwise, continue as normal.

This commit also edits the funding tab template, so subcommunities without a funding check configured display a message instead of showing up empty.

### DIFF
--- a/site/zenodo_rdm/checks/funding.py
+++ b/site/zenodo_rdm/checks/funding.py
@@ -8,11 +8,15 @@
 import hashlib
 import json
 
+from invenio_access.permissions import system_identity
 from invenio_checks.base import Check, CheckResult
 from invenio_checks.models import CheckConfig
 from invenio_communities.proxies import current_communities
 from invenio_pidstore.errors import PIDDoesNotExistError
+from invenio_rdm_records.checks.requests import CommunityInclusion, CommunitySubmission
 from invenio_records_resources.proxies import current_service_registry
+from invenio_requests.proxies import current_requests_service
+from invenio_search.api import dsl
 
 from zenodo_rdm.orcha.utils import run_funding_relevance_workflow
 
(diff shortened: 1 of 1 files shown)


templates/semantic-ui/invenio_checks/requests/funding_tab.html
mairasalazar, line 1: If the `filter_by_community` changes ([zenodo-rdm](https://github.com/zenodo/zenodo-rdm/pull/1418) and [invenio-checks](https://github.com/inveniosoftware/invenio-checks/pull/86)) are accepted, there is no need for these changes

site/zenodo_rdm/checks/funding.py
```diff
@@ -125,6 +151,13 @@ def get_updated_result(check_result, message, success):
         community = current_communities.service.record_cls.get_record(config.community_id)
         award_descriptions = self._get_awards_description(record, community)
 
+        if community.slug == 'eu' and self._get_ec_requests(record, community).total == 0:
```
slint:
```suggestion
        if community.slug == "eu" and self._get_ec_requests(record, community).total == 0:
```

nit: or `ruff` eventually will complain ;)

slint: nit: I would extract these boolean checks into separate vars before the `if ...:` for better readability.

slint (approved):

libcairo.2.dylib
```diff
@@ -0,0 +1 @@
+/opt/homebrew/lib/libcairo.2.dylib
```
slint: *L i b C a i r o*  
*l e a k*
😱

mairasalazar: 🙀 sorry, fixed!

PR #1419, opened 2026-08-06
feat(checks): add community member link to check

### LINKED ISSUE #1420: Improvements HZ+ checks

mairasalazar:
User names in the subcommunity member check tab now link to either the profile settings page for the user, or to the administration user page for admins. Adds a ZenodoChecks Flask extension to expose administration_permission as a Jinja global.

### DIFF
--- a/site/zenodo_rdm/theme/ext.py
+++ b/site/zenodo_rdm/theme/ext.py
@@ -1,0 +1,19 @@
+# SPDX-FileCopyrightText: 2026 CERN
+# SPDX-License-Identifier: GPL-3.0-or-later
+"""ZenodoRDM Theme module."""
+
+from invenio_administration.permissions import administration_permission
+
+
+class ZenodoTheme:
+    """Zenodo Theme extension."""
+
+    def __init__(self, app=None):
+        """Extension initialization."""
+        if app:
+            self.init_app(app)
+
+    def init_app(self, app):
+        """Flask application initialization."""
+        app.extensions["zenodo-theme"] = self
+        app.jinja_env.globals["is_admin"] = administration_permission.can
--- a/templates/semantic-ui/invenio_checks/requests/subcommunity_member_tab.html
+++ b/templates/semantic-ui/invenio_checks/requests/subcommunity_member_tab.html
@@ -28,9 +28,20 @@ under the terms of the MIT License; see LICENSE file for more details.
             <div class="rel-pt-1 rel-pb-1">
               {% for value in check_result.value %}
                 <div class="ui {{ value.level }} small message">
-                  {% if value.name %}<strong>{{ value.name ~ " - " ~ value.role }}</strong>
+                  {% set is_self = current_user.is_authenticated and current_user.id == value.id %}
+                  {% if value.name %}
+                    <strong>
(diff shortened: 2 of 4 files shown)


site/zenodo_rdm/checks/ext.py
```diff
@@ -0,0 +1,19 @@
...
+
+    def __init__(self, app=None):
+        """Extension initialization."""
+        if app:
+            self.init_app(app)
+
+    def init_app(self, app):
+        """Flask application initialization."""
+        app.extensions["zenodo-checks"] = self
+        app.jinja_env.globals["administration_permission"] = administration_permission
```
mairasalazar: Maybe overkill to register an extension just for this, but since we'll continue expanding zenodo checks (e.g. metadata and content comparison checks), I went with this approach

slint: tbh, I feel this should have been in Invenio-Administration for a long time now. Our concern is that the REST API and UI should always return the same info, but in reality we've missed out on providing a better UX (for admins, often) because of this.

minor: could we expose a higher-level `is_admin()` Jinja global instead? It could simply call `administration_permission.can()`, allowing templates to use it like `{% if is_admin() %}` without exposing the underlying permission object.

templates/semantic-ui/invenio_checks/requests/subcommunity_member_tab.html
```diff
@@ -28,9 +29,20 @@
...
               {% for value in check_result.value %}
                 <div class="ui {{ value.level }} small message">
-                  {% if value.name %}<strong>{{ value.name ~ " - " ~ value.role }}</strong>
+                  {% set is_self = current_user.is_authenticated and current_user.id == value.id %}
+                  {% if value.name %}
+                    <strong>
+                      {% if is_self %}
+                        <a href="/account/settings/profile">{{ value.name }}</a>
+                      {% elif is_admin %}
+                        <a href="/administration/users?q=id%3A{{ value.id }}">{{ value.name }}</a>
```
slint: minor: I know it'll might be painful, but we should use `invenio_url_for` for generating these URLs instead of hardcoding. I know that in JS we do this often (since there's no `invenioUrlFor` to call), but while in Jinja we have quite some flexibility.

slint (commented):
LGTM in principle, two comments on how the Jinja global could be implemented + using `invenio_url_for`

site/zenodo_rdm/theme/ext.py
slint, line 1: This can be moved under `site/zenodo_rdm/theme/`

slint (approved):

PR #1424, opened 2026-08-11
refactor(exporter): reorganize record export flow

slint:
> AI disclosure: I directed this refactor with Codex and reviewed the resulting diff in detail. AI wrote pretty much most of the move, new implementations, and tests. I had to go through many iterations though, referencing [the legacy Zenodo exporter module](https://github.com/zenodo/zenodo/tree/master/zenodo/modules/exporter), and getting the right naming and "split" between the different concepts.

* Split the logic from the Celery task into separate files. Some part of
  the design was borrowed from the legacy Zenodo exporter module.
* Produce multiple export formats from one record scan.
* Use a temporary "staging" file(s) to write the archive(s) before creating
  the final published ObjectVersion(s).


scripts/export-all-xml.py
```diff
@@ -10,18 +10,20 @@
 """
 
 from invenio_app.factory import create_api
+
 from zenodo_rdm.exporter.tasks import export_records
 
 
 def export_all_xml():
```
ptamarit: We should get rid of `xml` in the filename and method name.

ptamarit: We should remove the "xml" from this filename and method name.

PR #1431, opened 2026-08-21
Switch to Python 3.14 and pnpm + rspack

slint:
- **feat: Python 3.14 support**
  Run the app on Python 3.14. No Invenio base image exists for 3.14 yet, so
  this adds a single-stage AlmaLinux Dockerfile that sets up Python (via uv),
  Node.js + pnpm and the system libraries from scratch.
  
  * pyproject/uv.lock: requires-python >=3.14; pyuwsgi -> uwsgi; add sword2
    (slint fork @support-314) and invenio-xrootd; lxml -> 6.1.1 (has cp314 wheels,
    4.9.4 fails to compile on 3.14)
  * XRootD: install the self-contained cp314 wheel (xrootd 5.9.6); no system
    XRootD RPMs since the manylinux wheel bundles the client libs
  * Dockerfile: single-stage build on almalinux:9; assets built with
    `invenio webpack buildall`; drop the assets tree afterwards
  * add .dockerignore (excludes .git, node_modules, *.db, migrator/, var/, tmp/)
  * CI: run tests on Python 3.14
  

- **feat(assets): switch to pnpm and rspack**
  Replace the npm/webpack asset toolchain with pnpm + rspack, matching upstream
  Invenio's default builder.
  
  * commit root package.json + pnpm-lock.yaml (rspack, pnpm 11.15.1)
  * add a site/ lint workspace and move eslint/prettier config into it
  * invenio.cfg: use the rspack webpack project and pynpm PNPMPackage
  * run-js-linter.sh and CI drive linting through pnpm; README documents the flow
  * .invenio: javascript_package_manager = pnpm
  

- **release: v26.0.0**
  

- **fix(deps): support current Sentry and XRootD**
  * Use invenio-logging's compatibility transport because CERN Sentry 9 does not accept envelope events.
  * Keep XRootD in the locked Python 3.14 runtime extras instead of installing it separately in the image.
  

- **build(docker): optimize Python 3.14 image**
  * Build on the maintained Invenio base instead of recreating its runtime on AlmaLinux.
  * Export stable dependency-only inputs so release version bumps and source edits do not invalidate expensive Python and pnpm install layers.


.github/workflows/tests.yml
```diff
@@ -71,11 +71,17 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v7
 
+      - name: Install pnpm
+        uses: pnpm/action-setup@v4
```
ptamarit: It looks like there is a a new action called `pnpm/setup`:

https://github.com/pnpm/action-setup#migrating-to-pnpmsetup

pyproject.toml
```diff
@@ -33,6 +34,9 @@ invenio-swh = { git = "https://github.com/inveniosoftware/invenio-swh", rev = "v
 invenio-records-resources = { git = "https://github.com/inveniosoftware/invenio-records-resources", branch = "fix-read-many" }
 invenio-app-rdm = { git = "https://github.com/inveniosoftware/invenio-app-rdm", branch = "feature/orcha" }
 invenio-rdm-records = { git = "https://github.com/inveniosoftware/invenio-rdm-records", branch = "feature/orcha" }
+sword2 = { git = "https://github.com/slint/python-client-sword2", branch = "support-314" }
```
ptamarit: Wouldn't it be better to fix this in `invenio-swh` as discussed here inveniosoftware/invenio-swh#49 ?
Would it make sense to host this fork on the `inveniosoftware` org?

ptamarit (approved):
Disclaimer: I reviewed the code but did not test locally.

.github/workflows/tests.yml
slint: Nice, looks like we can also install Node.js from it, so one less step in the workflow ;)

pyproject.toml
slint: Agree, I think it's time to fix this a bit:

- I've forked sword2 over to `inveniosoftware` and pushed my branch to `master`: https://github.com/inveniosoftware/python-client-sword2
- For `invenio-swh` I think we need to first merge https://github.com/inveniosoftware/invenio-swh/pull/51 and then rebase https://github.com/inveniosoftware/invenio-swh/pull/49 on top, so that we can reference our fork in the dependencies (needs to be in `pyproject.toml` which supports git dependencies like `sword2 @ git+https://github.com/inveniosoftware/python-client-sword2@0.3.1`). I, *or actually the clanker* 😅, will have a go at it since `setup.cfg` was rewritten and there's modification in your PR.

OliverGeneser (approved):
Tested locally and works great!

PR #1432, opened 2026-08-24
Fix exporter error handling and logging

slint:
- **fix(exporter): continue after malformed records**
  * Preserve usable formats and subsequent records when one serialization fails.
  * Keep archive and storage failures fatal so incomplete output is not silently published.
  

- **feat(exporter): log progress and throughput**
  * Export runs can take hours, so operators need an accurate total, processing rate, and ETA rather than a debug-only record index.
  * The final summary exposes format-level errors without treating data-specific failures as a failed run.
  

- **fix(exporter): preserve Sentry error grouping**
  * Use lazy logging arguments so record-specific values do not split related failures into separate Sentry issues.

### DIFF
--- a/site/tests/exporter/test_writers.py
+++ b/site/tests/exporter/test_writers.py
@@ -67,3 +67,44 @@ def test_write_consumes_records_once_for_all_outputs(app, tmp_path):
     assert iterations == ["started"]
     assert _tar_members(paths[0]) == {}
     assert _tar_members(paths[1]) == {}
+
+
+def test_write_continues_after_record_errors(app, tmp_path, minimal_record, caplog):
+    """Malformed records do not stop other records or formats."""
+    invalid_xml = dict(
+        minimal_record,
+        id="invalid-xml",
+        metadata=dict(
+            minimal_record["metadata"],
+            funding=[
+                {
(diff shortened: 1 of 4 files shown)


OliverGeneser (commented):
Hmm from what I can see we previously raised on errors, so a failed export wouldn't upload or prune anything. But now we just drop the failing records from the archive, so we end up publishing incomplete ones. Is this the new intended behavior? 😊    
Also from what I see the errors never leaves write_archives, so it isn't tracked in the task/job but only as a log trace?

slint:
> We previously raised on errors... Is this the new intended behavior?

The original behavior actually (before the multi-format writer) was to skip errors and log them in the output, since a few problematic records shouldn't fail the entire export. Then we could look into the records from the logs and potentially fix them for the next dump.

> Also from what I see the errors never leaves write_archives, so it isn't tracked in the task/job but only as a log trace?

We report a "final" error count with the last call to `_log_progress`, and there's not much more info to return to the caller (e.g. stacktraces or record IDs). I think the better option would be to fully log errors in a separate file, so that we also don't end up keeping things in memory... I'm not super eager about adding this now, I'll have to think a bit about where in the "stream" this would have to go.

PR #1433, opened 2026-08-25
build(docker): optimize final built image size

slint:
* Split into buider and runtime stages.
* Set both `PNPM_CONFIG_STORE_DIR` and `PNPM_CONFIG_CACHE_DIR`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,20 +2,20 @@
 #
 # Dockerfile that builds a fully functional image of the Zenodo app.
 #
-# The Invenio base image provides Python 3.14, Node.js, pnpm, uv and the common
-# system libraries and instance layout used by Invenio applications.
-#
-# Build:
-#   docker build -t zenodo:latest .
+# The Invenio base image is used to build the app, but the final image is based on
+# `python:3.14-slim-trixie` to reduce the image size.
 
 ARG BUILD_PLATFORM=linux/amd64
 ARG BUILD_EXTRAS="--extra sentry --extra xrootd"
 
 FROM --platform=${BUILD_PLATFORM} ghcr.io/inveniosoftware/invenio:debian-python3.14 AS base
 
+ENV PNPM_CONFIG_STORE_DIR=/pnpm/store \
+    PNPM_CONFIG_CACHE_DIR=/pnpm/cache
+
 # --- Python dependencies ---
-# Export dependencies separately from the application packages. Version-only
-# changes produce the same output, allowing BuildKit to reuse the install layer.
+# Install dependencies separately from a requirements.txt export, to avoid busting the
+# cache when we bump the app version.
 FROM base AS python-requirements
 ARG BUILD_EXTRAS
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -2,20 +2,20 @@
...
+# The Invenio base image is used to build the app, but the final image is based on
+# `python:3.14-slim-trixie` to reduce the image size.
 
 ARG BUILD_PLATFORM=linux/amd64
 ARG BUILD_EXTRAS="--extra sentry --extra xrootd"
 
 FROM --platform=${BUILD_PLATFORM} ghcr.io/inveniosoftware/invenio:debian-python3.14 AS base
 
+ENV PNPM_CONFIG_STORE_DIR=/pnpm/store \
+    PNPM_CONFIG_CACHE_DIR=/pnpm/cache
```
slint, line 14: Both of these had to be set (and the paths used as cache volumes) to avoid `pnpm ...` commands re-downloading packages.

Dockerfile
```diff
@@ -87,37 +81,94 @@ COPY ./translations ${INVENIO_INSTANCE_PATH}/translations
...
 
+# The instance must be group-writable, since the container can run under an arbitrary
+# UID in GID 0. Done here rather than in the final image, where a recursive chown would
+# duplicate the whole tree into a second layer; COPY --from preserves owner and mode.
+RUN chown -R invenio:0 ${WORKING_DIR} && chmod -R g=u ${WORKING_DIR}
+
+# --- Application image ---
+# Must stay on the same Python and Debian release as the base image above, so that the
+# venv and its compiled extensions still match.
+FROM --platform=${BUILD_PLATFORM} python:3.14-slim-trixie
```
slint, line 115: For the "runtime" image, we can use the slim version, since we're not building/compiling anything.

OliverGeneser (commented):
Is libcairo2 provided in the slim-downed image? Because I remember needing to add the to my nix flake 😆

ntarocco (approved):
LGTM!
Did you test files access and krb?

slint:
> LGTM! Did you test files access and krb?

The previous version deployed on Sandbox is not much different in terms of these parts, and things seem to work nicely, i.e. file (re)uploads are good and downloads go through the EOS HTTP gateway. It's worth seeing if the slim final image makes any difference, but I verified locally the Python imports, reading some public files on EOS public. I think it's "cheaper" to just test directly on Sandbox at this point 😅

slint:
> Is libcairo2 provided in the slim-downed image? Because I remember needing to add the to my nix flake 😆

Pushed a fix right now for this as well, so that we explicitly install libcairo. I found a weird bug as part of this on Sandbox that is now patched and released at https://github.com/inveniosoftware/invenio-formatter/pull/138 . In general I'm not happy with how complex the badges code is, so I think there might be a way to simplify and remove the whole image-processing parts from `invenio-formatter` eventually, and thus simplify a bit the lib-dependencies we require in the Docker images.

PR #1444, opened 2026-09-03
fix: prevent hyperlink injection in zenodo email

ciurana-life:
Closes: https://github.com/zenodo/ops/issues/537

* Full name allows characters like `.` in them making it possible to make a url your `full name`
* Removing the `full name` from the email prevents vulnerabilities.

### DIFF
--- a/templates/security/email/change_notice.html
+++ b/templates/security/email/change_notice.html
@@ -8,9 +8,8 @@
             </td>
         </tr>
         <tr>
-            {% set name = user.profile.full_name if user.profile and user.profile.full_name else user.username %}
             <td style="padding:5px 15px 10px 15px;">
-                <p>{{ _('Hello, %(name)s!', name=name) }}</p>
+                <p>{{ _('Hello, @%(username)s!', username=user.username) }}</p>
                 <p>{{ _('Your password has been successfully changed!') }}</p>
             </td>
         </tr>
--- a/templates/security/email/welcome.html
+++ b/templates/security/email/welcome.html
@@ -9,7 +9,7 @@
         </tr>
         <tr>
             <td style="padding:5px 15px 10px 15px;">
-                <p>{{ _('Welcome %(username)s!', username= (user.user_profile.get('full_name') or user.username)) }}</p>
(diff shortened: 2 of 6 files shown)


templates/security/email/welcome.html
```diff
@@ -9,7 +9,7 @@
         </tr>
         <tr>
             <td style="padding:5px 15px 10px 15px;">
-                <p>{{ _('Welcome %(username)s!', username= (user.user_profile.get('full_name') or user.username)) }}</p>
+                <p>{{ _('Welcome %(username)s!', username=user.username) }}</p>
```
slint:
```suggestion
                <p>{{ _('Welcome @%(username)s!', username=user.username) }}</p>
```

nit: to mimic a bit how GitHub does it, and also make it a bit more clear that this is a username.

templates/security/email/welcome.html
slint, line 1: Do we have other emails that could be triggered and greet with the full name?

slint (approved):
LGTM, a nit and a question on other possible templates

ntarocco (approved):

templates/security/email/welcome.html
ciurana-life: done

templates/security/email/welcome.html
ciurana-life: might be they are a problem at some point, did the same change on all of them

ntarocco (approved):

templates/security/email/change_notice.html
```diff
@@ -8,9 +8,8 @@
             </td>
         </tr>
         <tr>
-            {% set name = user.profile.full_name if user.profile and user.profile.full_name else user.username %}
             <td style="padding:5px 15px 10px 15px;">
-                <p>{{ _('Hello, %(name)s!', name=name) }}</p>
+                <p>{{ _('Hello, @%(username)s!', username=user.username) }}</p>
```
slint, line 12: Sorry, one last minor thing I just realised. There are cases in production with users that don't have a username. Can we default then to just "Zenodo user"?

slint (approved):

PR #1445, opened 2026-09-04
fix: emailing users without usernames

ciurana-life:
* full name was removed from emails because of a vulnerability
* we replaced with username
* some accounts in Zenodo don't have username

### DIFF
--- a/templates/security/email/change_notice.html
+++ b/templates/security/email/change_notice.html
@@ -9,7 +9,8 @@
         </tr>
         <tr>
             <td style="padding:5px 15px 10px 15px;">
-                <p>{{ _('Hello, @%(username)s!', username=user.username) }}</p>
+                {% set name = "@" ~ user.username if user.username is not none else "Zenodo user" %}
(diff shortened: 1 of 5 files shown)


templates/security/email/change_notice.html
```diff
@@ -9,7 +9,8 @@
         </tr>
         <tr>
             <td style="padding:5px 15px 10px 15px;">
-                <p>{{ _('Hello, @%(username)s!', username=user.username) }}</p>
+                {% set name = "Zenodo user" if user.username is none else "@" ~ user.username %}
```
slint:
```suggestion
                {% set name = "@" ~ user.username if user.username is not none else "Zenodo user" %}
```

turbo-nit: just to have the "green path" first in the conditional

slint (approved):
LGTM! One turbo-nit on the style of the conditional

templates/security/email/change_notice.html
ciurana-life: done!

PR #1452, opened 2026-09-15
feat: benchmark write and read endpoints

ciurana-life:


benchmark/common.py
```diff
@@ -0,0 +1,162 @@
...
+    cached = cache.get(email)
+
+    if cached:
+        requests.utils.cookiejar_from_dict(cached["cookies"], cookiejar=client.cookies)
+        check = client.get(
+            "/account/settings/profile",
+            name="/account/settings/profile [session check]",
+        )
+        if check.status_code == 200:
+            return {"X-CSRFToken": cached["csrf_token"]}
```
OliverGeneser, line 107: In case the session expired wouldn't this then redirect and still give a status 200?

benchmark/provision_accounts.sh
```diff
@@ -0,0 +1,27 @@
...
+
+echo "email,password" > "${OUT_FILE}"
+
+for ((i = 1;i <= COUNT; i++)); do
+    email="locust-${i}@demo.org"
+    password="$(LC_ALL=C tr -dc 'A-Za-z0-9' </dev/urandom | head -c 24)"
+    echo "Creagin ${email}..."
+    ${RUN_INVENIO} users create "${email}" --password "${password}" --active --confirm
+    echo "Created ${email}, ${password}"
+    echo "${email},${password}" >> "${OUT_FILE}"
```
OliverGeneser: nit; this would still save the users if the creation fails

benchmark/provision_accounts.sh
```diff
@@ -0,0 +1,27 @@
...
+# accounts for the Loucst write/authenticated-read scenarios.
+#
+# Usage:
+#    ./provision_accounts.sh [num_accounts]
+#
+# Writes email, password pairs to benchmark/accounts.csv (gitignored)
+# TODO: Pipe it to dev or sandbox 
+
+COUNT="${1:-10}"
+OUT_FILE="accounts.csv"
```
OliverGeneser: nit; maybe this should be relative, so we can run the benchmark from the root folder?

benchmark/reads.py
```diff
@@ -0,0 +1,108 @@
...
+    @task(10)
+    def get_logos(self):
+        if not hasattr(self, "logos"):
+            response = self.client.get("/api/communities")
+            self.logos = [
+                hit["links"]["logo"] for hit in response.json()["hits"]["hits"]
+            ]
+        self.client.get(
+            urlparse(random.choice(self.logos)).path,
+            name=f"[{self.label}] /api/communities/[random id]/logo",
```
OliverGeneser, line 89: note; this gives a 404 for all communities without a logo

benchmark/common.py
```diff
@@ -0,0 +1,170 @@
...
+    "small": (200 * 1024, "paper.pdf"),
+    "medium": (5 * 1024 * 1024, "dataset.csv"),  # 5.2mb
+    # "large": (20 * 100 * 1024 * 1024, "big-dataset.zip") # 2gb
+    "large": (100 * 1024 * 1024, "big-dataset.zip"),
+}
+
+
+def random_file(size_tier: str) -> tuple[str, bytes]:
+    size, filename = _SIZE_TIERS[size_tier]
+    return f"{uuid.uuid4()}_{filename}", os.urandom(size)
```
OliverGeneser: Hmm I think we should only generate the random data once and then just generate a new name each time 🤔

OliverGeneser (approved):
The session-cache is currently a bit iffy as it's not atomic and erases for every account, but I think it's fine for now 😊

benchmark/reads.py
ciurana-life: Is fine since it is reproducing the real behavior.

benchmark/common.py
ciurana-life: done!

